PR #1, opened 2024-07-25
SYS-1658: Docker environment for ArhivesSpace development

ztucker4:
Implements [SYS-1658](https://uclalibrary.atlassian.net/browse/SYS-1658)

Adds a `docker-compose.yml` file, and other supporting files, to allow for Docker-based use of ArchivesSpace for local development.

As described in the README, build and run the containers with `docker-compose`:
```
docker-compose build
docker-compose up -d
```

Once everything is up and running, the staff interface will be available at [http://localhost:8080](http://localhost:8080) (public interface at [http://localhost:8081](http://localhost:8081)). Log in with username and password `admin`. 

Four containers are created: `as_db`, `as_solr`, `as_archivesspace`, and `as_python`. The configuration for the Solr and DB containers was largely copied from the ArchivesSpace repo - specifically, I modified the [docker-compose file](https://github.com/archivesspace/archivesspace/blob/f70d672f5648f4da7cdca471fe3745431467ca52/docker-compose.yml) and used their [Solr Dockerfile](https://github.com/archivesspace/archivesspace/blob/f70d672f5648f4da7cdca471fe3745431467ca52/solr/Dockerfile) and related [Solr config files](https://github.com/archivesspace/archivesspace/tree/f70d672f5648f4da7cdca471fe3745431467ca52/solr) in their entirety. The Python container currently does nothing and has only `archivessnake` installed from `requirements.txt`.

I have a couple of questions/unresolved issues about this PR:
1. In order to prevent an error about mismatched checksums for the Solr schema, I've set `APPCONFIG_SOLR_VERIFY_CHECKSUMS="false"` in the ArchiveSpace container's `.env` file. I'm not sure if the problem might be because I'm using a slightly older version of the `solr` and `archivesspace` docker images from Docker Hub (3.5.0 vs. current 3.5.1), but I took the `solr/schema.xml` file from the [current ArchivesSpace repo](https://github.com/archivesspace/archivesspace/blob/master/solr/schema.xml). I don't think this matters much for local development though, as searching appears to work even with the check disabled.
2. The ArchivesSpace container is really slow on my machine. Building only takes a few seconds, but running all of the setup (mainly migrations and indexing, per the parts of the container's logs that I can understand) takes 20+ minutes for me. Each page load in browser takes 15-30 seconds once everything is running. This may be because I have an M1 chip and the image is amd64.

### DIFF
--- a/.docker-compose_archivesspace.env
+++ b/.docker-compose_archivesspace.env
@@ -1,0 +1,10 @@
+APPCONFIG_DB_URL="jdbc:mysql://db:3306/archivesspace?useUnicode=true&characterEncoding=UTF-8&user=as&password=as123"
+# disable the embedded solr
+APPCONFIG_ENABLE_SOLR="false"
+APPCONFIG_SOLR_URL="http://solr:8983/solr/archivesspace"
+APPCONFIG_FRONTEND_PROXY_URL="http://localhost:8080"
+APPCONFIG_PUBLIC_PROXY_URL="http://localhost:8081"
+# run DB migrations to get working empty database
+ASPACE_DB_MIGRATE="true"
+# temporary(?) - disable Solr schema checksums
+APPCONFIG_SOLR_VERIFY_CHECKSUMS="false"
--- a/.docker-compose_db.env
+++ b/.docker-compose_db.env
@@ -1,0 +1,5 @@
+# database config
+MYSQL_ROOT_PASSWORD="123456"
+MYSQL_DATABASE="archivesspace"
+MYSQL_USER="as"
+MYSQL_PASSWORD="as123"
--- a/README.md
+++ b/README.md
@@ -1,2 +1,17 @@
 # archivesspace-toolkit
 An environment for local development and testing of tools to update UCLA's ArchivesSpace records.
+
+## Local Setup
+Build and run the containers using docker-compose.
+
+`docker-compose build`
+`docker-compose up -d`
+
+Wait until you see the below message in the `as_aspace` container logs:
+```
+Welcome to ArchivesSpace!
+You can now point your browser to http://localhost:8080
+```
+
+The staff interface will be available at `http://localhost:8080`, and the public interface will be at `http://localhost:8081`. Log in with username and password `admin`.
+
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,38 @@
+services:
+  python:
+    container_name: as_python
+    build: 
+      context: ./python
+
+  archivesspace:
+    container_name: as_aspace
+    image: archivesspace/archivesspace:3.5.0
+    ports:
+      - "8080:8080"
+      - "8081:8081"
+      - "8082:8082"
+      - "8089:8089"
+      - "8090:8090"
+    depends_on:
+      - db
+      - solr
+    env_file:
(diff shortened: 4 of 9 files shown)


ztucker4:
Hi @akohler, I've made a few changes on this PR and have some notes:
* I've added a `healthcheck` to the DB container. Now the main ArchivesSpace container should wait until the DB container is ready to receive connections before starting up (checked with a simple `SHOW DATABASES;` command). The timeout values used here consistently work for me - after the initial 20s wait, the DB is up and running and the AS container starts happily.
  * To support this, I moved the DB container's environment variables out of `.docker-compose_db.env` and into the `docker-compose` file directly. I could not figure out how to access these variables in the `healthcheck` command if they were defined in the external `.env` file. I think this is probably ok since this is all for a development environment, but if you know how to fix this it might be good to know.
* For performance of the AS container, I had success building the image from scratch locally. Pages now load much faster and total startup time is <3 minutes. I wasn't able to make any improvements to performance of the prebuilt image, but the local image should work for now.

akohler (changes requested):
@ztucker4 The new `healthcheck` works well for me, thank you.  However, it should not be necessary to add the db variables inline - please see my separate comment for code which works for me.

I'm glad using a local build works for you, and it's fine with me to stick with that.  Since this is a dependency (must pull and build a separate repo), please document it briefly with a comment in `docker-compose.yml` (explaining where `archivespace-local` comes from) and in more detail in `README.md` (how to build and tag that image as required for this system).

Finally, we'll want permanent storage for the database and probably for solr too, so we don't have to wait through migrations and indexing - especially after we import the production database dump. Please add volumes for both.  For an example of multiple services with volumes, see https://github.com/UCLALibrary/oral-history/blob/main/docker-compose.yml

Thanks --Andy

ztucker4:
With today's changes (and using the latest version of the `archivesspace` repo to build the AS image), this now works consistently for me with persistent volumes for both the `solr` and `db` containers. Since standup, I've also added the missing `.docker-compose_db.env` file to this PR and updated `README.md` with brief build instructions.

akohler (approved):
@ztucker4 This all looks great now.  I'm fine with leaving `APPCONFIG_SOLR_VERIFY_CHECKSUMS="false"` in our local config - seems compatible with [their docs](https://archivesspace.github.io/tech-docs/provisioning/solr.html), and we'll deal with problems if any relevant ones occur.

I made a few minor changes:
* Removed the explicit `image` from the `solr` service, because we're building locally, not pulling.  The resulting image just gets tagged as `archivesspace-toolkit-solr:latest`, which is appropriate since we'll be rebuilding as needed from local code.
* Made our local `archivesspace-toolkit/solr` directory identical to the official source, `archivesspace/solr`.  You had skipped `delete.sh` (which might be useful), and `.dockerignore` (which is like `.gitignore`, but specifies what local files to leave out of a docker image when built).  Definitely not critical, but since we're going rogue I want to stay identical to the original whenever practical.
* Fixed a typo in the build instructions, missing the path value  `.` for current directory. `docker build` requires a path.
* Added newline to the end of several files which lacked them.  Generally does not affect functionality, but having a newline makes viewing text files cleaner (especially via `cat`), and gets rid of annoying "No EOL" warnings in `git` and the corresponding red icons in Github.

Merged!

PR #4, opened 2024-08-14
SYS-1664: aspace user in Dockerfile, add ArchivesSnake config

ztucker4:
Implements [SYS-1664](https://uclalibrary.atlassian.net/browse/SYS-1664)

Based on the [batch CD project Dockerfile](https://github.com/UCLALibrary/music-cd-batch/blob/main/Dockerfile), a few changes to the python service's Dockerfile:

* Create and use a specific `aspace` user, rather than `root`.
* For consistency with the batch CD example, use a different base image (`python:3.11-slim-bullseye` instead of `python:3.11`), and explicitly set timezone.
* Copy the `.archivessnake.yml` config file (also included in this PR) into the `aspace` user's home directory, as expected by ArchivesSnake.

To test, rebuild the python service (`docker compose build python`) and observe changes in the rebuilt container.

[SYS-1664]: https://uclalibrary.atlassian.net/browse/SYS-1664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python/.archivessnake.yml
```diff
@@ -0,0 +1,5 @@
+baseurl: http://archivesspace:8089
+username: admin
+password: admin
+logging_config:
+default_config: DEBUG_TO_FILE
```
akohler, line 5: The [documentation](https://github.com/archivesspace-labs/ArchivesSnake?tab=readme-ov-file#configuration) does not make this very clear... but I believe the `default_config:` line needs to be indented, because it's a
> Hash with various config values for the logging subsystem

As-is, when I try to instantiate this, I get this error (this is with the local system fully started, and the `admin` password set to `admin`):
```
$ docker compose run python python
Python 3.11.9 (main, Aug 13 2024, 01:17:41) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.

>>> from asnake.aspace import ASpace
>>> a = ASpace()

  File "<stdin>", line 1, in <module>
  File "/home/aspace/.local/lib/python3.11/site-packages/asnake/aspace/__init__.py", line 15, in __init__
    self.client = ASnakeClient(**config)
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aspace/.local/lib/python3.11/site-packages/asnake/client/web_client.py", line 73, in __init__
    if 'default_config' in self.config['logging_config']:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
```

As-is, the `logging_config` key has no value assigned (`None`), because there's nothing else on that line in the yaml file, and nothing indented below it to indicate it has any data.

When I indent line 5 by 2 spaces, I'm able to instantiate without errors, and can connect to my local instance.

akohler (changes requested):
@ztucker4 In general this is good: I can build and run the container, permissions are all correctly set, contents are as expected.

There's one problem with the `.archivessnake.yml` structure which prevents me from starting an instance of `asnake.ASpace()` - see detailed comment for that.

Please also add empty newlines at the end of both files - see my [comment on previous PR](https://github.com/UCLALibrary/archivesspace-toolkit/pull/1#pullrequestreview-2226462132) for more.

Finally, please add a bind mount volume to the `python` service via `docker-compose.yml`, to support live editing, similar to https://github.com/UCLALibrary/music-cd-batch/blob/main/docker-compose.yml .  That way, all changes in our local `python` directory will immediately be available in the container's `app` directory, which will make writing and testing code much easier.

Thanks --Andy

ztucker4:
Requested changes are done - files end with newlines, bind mount volume is added to `python` service in `docker-compose.yml`, and `.archivessnake.yml` is correctly indented. I wasn't able to replicate the issue you saw with the unindented version, but it works the same for me with the indentation added.

akohler (approved):

akohler:
@ztucker4 All looks great now.  I rebuilt & restarted, and was able to create and run a test program successfully within the container:
```
# python/test.py
from asnake.aspace import ASpace


def main():
    aspace = ASpace()
    for repo in aspace.repositories:
        print(repo.name)


if __name__ == "__main__":
    main()
```

```
$ docker compose run python python test.py
UCLA Library Special Collections
Test repository
```

All merged, thanks.

PR #5, opened 2024-08-15
SYS-1665: Script to find unlinked top containers for deletion

ztucker4:
Implements [SYS-1665](https://uclalibrary.atlassian.net/browse/SYS-1665)

A script to identify unlinked top-level containers for deletion. The script takes no arguments. It iterates through all top-level containers in the main repository (ID `2`) and outputs a file containing URIs for containers not linked to any objects. Running against a database recently restored from the prod dump, there are more results found by this script than are reported in [Laney's spreadsheet](https://docs.google.com/spreadsheets/d/1t9HGF5uX0-JcnK3zJnfb7KOjq-n9RXHQG8sZTNnc8og) (6701 in the spreadsheet, 7588 found by this script).

The `top_containers` endpoint is paginated, with an apparent (undocumented but experimentally found) maximum page size of 250. Using this page size, the script took just over an hour to run in my environment.

[SYS-1665]: https://uclalibrary.atlassian.net/browse/SYS-1665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python/get_unlinked_top_containers.py
```diff
@@ -0,0 +1,39 @@
...
+client = ASnakeClient()
+
+
+def get_unlinked_top_containers():
+    all_top_containers = client.get(
+        "repositories/2/top_containers", params={"all_ids": True}
+    ).json()
+    total_top_containers = len(all_top_containers)
+    logger.info(f"Total top containers: {total_top_containers}")
+    pages = (total_top_containers // 250) + 1
```
akohler, line 16: The "magic number" of 250 gets used twice (here and line 23).  The [API docs](https://archivesspace.github.io/archivesspace/api/#paginated-endpoints) say "default set in AppConfig" - is that exposed via something we can set in `archivessnake.yml`?  

I also see the [low-level API](https://github.com/archivesspace-labs/ArchivesSnake?tab=readme-ov-file#low-level-api) mentions `get_paged` which seems to handle paginated data for us.

If neither of those is useful for us, then add an optional `pagesize` parameter to `get_unlinked_top_container()`, and set the default value there to 250.  That would also require adding `argparse` or similar for command-line argument support.

akohler (changes requested):
@ztucker4 This works well.  I think we'll need to investigate the discrepancy in the total numbers, but we can do that separately from this code.

One change requested, please see my detailed comment about page size and pagination.
Thanks --Andy

ztucker4:
I've updated the script. The logic is simplified by using `get_paged()` and performance is improved (total run time down to 30 minutes). The page size settings mentioned in the documentation are set in a `config.rb` file in the application container, and I don't see a way to modify with ArchivesSnake config, so the script now takes an optional `--page_size` argument, defaulted to 250.

akohler (changes requested):
@ztucker4 I like `get_paged()`, much cleaner code and logs.  I verified functionality by running with the default, and by specifying `--page_size 1000`.  Both returned the same 7588 rows, matching your count.  Both runs took roughly the same amount of time, around 30 minutes for me too.

Two more small changes, then I'll merge:
1. See separate comment about enforcing argument type via `parseargs`
2. Please add a new section at the bottom of `.gitignore` - I usually call it `# Exclude logs and other output`.  `.log` is already excluded, but add `.txt` and any other extensions you think we might use, so output files we create don't get listed by git.

Thanks --Andy

ztucker4:
I've added the `type` parameter to the `--page_size` argument and updated `.gitignore` to exclude `.txt` and `.csv` files - ready for another look!

akohler (approved):

akohler:
@ztucker4 Looks great, and thanks for the newline ending `.gitignore` :) - merged!

PR #6, opened 2024-08-16
SYS-1666: script to delete unlinked top containers

ztucker4:
Implements [SYS-1666](https://uclalibrary.atlassian.net/browse/SYS-1666)

A script that uses ArchivesSnake to delete unlinked top containers as identified by `get_unlinked_top_containers.py`. The script takes one argument: the path to a text file containing URIs for deletion. This file is expected to be in the format output by `get_unlinked_top_containers.py` (one URI per line).

The script checks each URI to confirm existence and unlinked status of each top container. If a URI in the file is malformed (e.g. `/repositories/2/top_containers/string`), or corresponds to a top container that doesn't exist or has links, the script will log this, skip the URI, and continue through the file. 

[SYS-1666]: https://uclalibrary.atlassian.net/browse/SYS-1666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/delete_unlinked_top_containers.py
+++ b/python/delete_unlinked_top_containers.py
@@ -10,19 +10,17 @@ client = ASnakeClient()
 
 def container_is_unlinked(top_container_uri: str) -> bool:
     top_container = client.get(top_container_uri).json()
-    if len(top_container["collection"]) == 0:
-        return True
-    return False
+    return len(top_container["collection"]) == 0
 
 
-def container_exists(top_container_uri: str) -> bool:
+def container_has_errors(top_container_uri: str) -> bool:
     top_container = client.get(top_container_uri).json()
     if "error" in top_container.keys():
         logger.error(
             f"Error retrieving top container {top_container_uri}: {top_container['error']}"
         )
-        return False
-    return True
+        return True
+    return False
 
 
 def delete_unlinked_top_containers(container_list_file: str):
@@ -34,13 +32,21 @@ def delete_unlinked_top_containers(container_list_file: str):
     deleted_count = 0
     skipped_uri_list = []
     for container in container_list:
-        if not container_exists(container):
-            # error message already logged in container_exists()
+        if container_has_errors(container):
+            # error message already logged in container_has_errors()
             skipped_uri_list.append(container)
             continue
         if container_is_unlinked(container):
             logger.info(f"Deleting unlinked top container: {container}")
-            client.delete(container)
+            delete_response = client.delete(container)
+            if delete_response.status_code == 403:
+                logger.error(
+                    f"Permission denied deleting top container {container}:"
+                    f"{delete_response.json()}"
+                )
+                raise PermissionError(
+                    f"Permission denied deleting top container {container}"
+                )
             deleted_count += 1
         else:
             logger.info(
--- a/python/delete_unlinked_top_containers.py
+++ b/python/delete_unlinked_top_containers.py
@@ -39,7 +39,12 @@ def delete_unlinked_top_containers(container_list_file: str):
         if container_is_unlinked(container):
             logger.info(f"Deleting unlinked top container: {container}")
             delete_response = client.delete(container)
-            if delete_response.status_code == 403:
+            status_code = delete_response.status_code
+            if status_code == 200:
+                # All OK
+                deleted_count += 1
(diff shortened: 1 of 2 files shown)


python/delete_unlinked_top_containers.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def container_is_unlinked(top_container_uri: str) -> bool:
+    top_container = client.get(top_container_uri).json()
+    if len(top_container["collection"]) == 0:
+        return True
+    return False
+
+
+def container_exists(top_container_uri: str) -> bool:
```
akohler, line 18: This seems more appropriate as a general error checker, because it doesn't really check whether the container exists - only that the response contains an error.  There probably are other error cases which don't represent the (non)existence of the container, but also indicate we should not try to delete it.

If all we want is an existence check, we don't need to look inside the payload, but can take advantage of the client basically being a wrapper around `requests`.
```
>>> top_container_uri = "/repositories/2/top_containers/99999999"
>>> response = client.get(top_container_uri)
>>> print(response)
<Response [404]>
>>> print(response.status_code)
404
```

I think there's more value in the generic error checking: we can use a more generic function in multiple scripts (refactoring it into a utility file when needed...).  So I recommend just renaming this function to better reflect its general "there's a problem" purpose.

python/delete_unlinked_top_containers.py
```diff
@@ -0,0 +1,62 @@
...
+
+logging.setup_logging(filename="archivessnake.log", level="INFO")
+# set label for custom logger - all output will be in archivessnake.log
+logger = logging.get_logger("delete_unlinked_top_containers")
+client = ASnakeClient()
+
+
+def container_is_unlinked(top_container_uri: str) -> bool:
+    top_container = client.get(top_container_uri).json()
+    if len(top_container["collection"]) == 0:
```
akohler, line 13: A more Pythonic way to do this:
```return len(top_container["collection"]) == 0```

Since there's no logging or complex logic here, I recommend the one-liner replacing the conditional.

File: python/delete_unlinked_top_containers.py [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+    deleted_count = 0
+    skipped_uri_list = []
+    for container in container_list:
+        if not container_exists(container):
+            # error message already logged in container_exists()
+            skipped_uri_list.append(container)
+            continue
+        if container_is_unlinked(container):
+            logger.info(f"Deleting unlinked top container: {container}")
+            client.delete(container)
```
akohler, line 43: What happens if this is run by a user who does not have permission to delete containers?  For example, our default accounts in production just have `view_repository`, I think.  This is unlikely to happen, and we haven't accounted for this in Alma API scripts... but we should start.

If the API or client here raise an exception, we should decide what to do.  I'm OK with not catching it and crashing immediately, if the error message is meaningful.  For example, if this program is run with a non-existing/unreadable file, the immediate `FileNotFoundError` is fine - the error is clear.

If the exception error message is not really clear about not having permission to delete, we should catch, log (and possibly also print) a useful message, and immediately exit.

If there's no exception, just an error embedded in a response similar to the payload of "container not found", and the program would therefore try and fail to delete every URI in the file... we should detect the problem, raise/create and log/print a meaningful exception of our own, and immediately exit.

akohler (changes requested):
@ztucker4 This is an excellent start.  The program runs as expected: I ran it with a sample file of 5 URIs (3 unlinked, 1 linked, 1 non-existent).  The 3 unlinked were deleted (and I confirmed they're gone), the other 2 were skipped and all log messages were correct.

I'm asking for a few changes: a minor bit of Pythonic pedantry, a semantic renaming, and more explicit handling of what happens if deletion cannot be done.

Thanks --Andy

akohler (approved):

akohler:
@ztucker4 Thanks for making the changes.  The `PermissionError` on 403 is good.  After thinking about it, I added code to capture unknown errors as well - just in case.  I also fixed the problem with multiple unconnected `archivessnake.yml` config files, by changing the `Dockerfile` - please rebuild that image locally to be sure you have that change.
Thanks --Andy

PR #7, opened 2024-10-09
SYS-1680: barcode update script with Tom Bradley mappings

ztucker4:
Implements [SYS-1680](https://uclalibrary.atlassian.net/browse/SYS-1680)

A generic script for adding Alma barcodes to ArchivesSpace top containers, plus matching logic specific to the Tom Bradley papers collection.

The script assumes the presence of an `alma_api_keys.py` file in the `python` folder, like that used by existing Alma API scripts.

The script takes six parameters:
1. `--alma_environment` - either `sandbox` or `production`. Used to select the appropriate API key.
2. `--bib_id` - a single Alma Bib ID.
3. `--holdings_id` - a single Alma Holdings ID.
4. `--resource_id` - a single ArchivesSpace resource ID (the collection to be barcoded).
5. `--profile` - the config file containing the `match_containers()` function specific to the collection. For now, the only available profile is `config.bradley`.
6. `--asnake_config` - the path to an ArchivesSnake .yml config file containing the ArchivesSpace URL and credentials.

A full example command, with data values corresponding to the actual Tom Bradley collection (in Sandbox Alma):
```
--alma_environment sandbox \
--bib_id 997094533606533 \
--holdings_id 22521149380006533 \
--resource_id 3828 \
--profile config.bradley \
--asnake_config .archivessnake.yml
```
Using a local ArchivesSpace instance, this full run took about 15 minutes for me. This updated 4399 barcodes, and found 102 unmatched Alma items and 10 unmatched ASpace top containers. Logs are written to `archivessnake.log`. Unmatched Alma and ASpace items (if any) are written to `unmatched_alma_items.json` and `unmatched_aspace_containers.json`. Top containers with existing barcodes are written to `top_containers_with_barcodes.json` and skipped in the matching process.

The specific matching logic for the Tom Bradley containers, defined in `bradley.py`, consists of splitting the Alma "Description" into two parts - "container type" and "indicator", separated by a period. Leading zeroes and trailing ` RESTRICTED` strings are removed from the indicator, and then these values are used to match against Top Containers under the specified Resource.

As the script uses the recently-created `alma_api_client` package, this PR includes the necessary changes to the Python service's Dockerfile and `requirements.txt`.

Also includes tests for `bradley.py`, with test JSON files containing actual API responses with some values changed. To run tests, within the `tests` folder in the container, run `python -m unittest test_bradley_mapping.py`.

[SYS-1680]: https://uclalibrary.atlassian.net/browse/SYS-1680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python/config/bradley.py
```diff
@@ -0,0 +1,45 @@
...
+        alma_container_type = description.split(".")[0]
+        alma_indicator = description.split(".")[1]
+        # if indicator starts with 0, remove it
+        while alma_indicator[0] == "0":
+            alma_indicator = alma_indicator[1:]
+        # if indicator ends with " RESTRICTED", remove it
+        if alma_indicator.endswith(" RESTRICTED"):
+            alma_indicator = alma_indicator[:-11]
+        # match with ASpace top container based on container type and indicator
+        for tc in aspace_containers:
```
akohler, line 17: As we talked about, this nested loop could be quite slow.  Instead, consider a method which builds a dictionary of ASPace data, keyed on a tuple of data needed by this config for matching Alma data, something like
```
def _get_aspace_match_data(list) -> dict[tuple]:
    # for each item
    # create "matchable" (tuple) with the ordered elements which are needed to match alma
```

python/config/bradley.py
```diff
@@ -0,0 +1,45 @@
+def match_containers(alma_items: list, aspace_containers: list, logger) -> tuple:
+    unmatched_alma_items = []
+    matched_aspace_containers = []
+    for item in alma_items:
+        item_id = item.get("item_data").get("pid")
```
akohler, line 5: Similar to the ASpace loop below, I think this could be a method which returns a dictionary of Alma data, keyed on a tuple of data needed by this config for matching ASpace data, something like:
```
def _get_alma_match_data(list) -> dict[tuple]:
    # for each item
    # create "matchable" (tuple) with the ordered elements which are needed to match aspace
```

python/config/bradley.py
```diff
@@ -0,0 +1,45 @@
...
+            alma_indicator = alma_indicator[1:]
+        # if indicator ends with " RESTRICTED", remove it
+        if alma_indicator.endswith(" RESTRICTED"):
+            alma_indicator = alma_indicator[:-11]
+        # match with ASpace top container based on container type and indicator
+        for tc in aspace_containers:
+            tc_type = tc.get("type")
+            tc_indicator = tc.get("indicator")
+            tc_uri = tc.get("uri")
+            if tc_type == alma_container_type and tc_indicator == alma_indicator:
```
akohler, line 21: Once there are 2 dictionaries of tuples for matching, this comparison can be more generic:
```
for key in aspace_keys:
    if key in alma_keys:
       # do something
```

python/config/bradley.py
```diff
@@ -0,0 +1,45 @@
...
+                # add barcode to top container
+                tc["barcode"] = barcode
+                matched_aspace_containers.append(tc)
+                break
+        else:
+            if logger:
+                logger.info(f"No match found for item {item_id}")
+            unmatched_alma_items.append(item)
+    # find unmatched ASpace top containers
+    unmatched_aspace_containers = []
```
akohler, line 33: Since we want to know about matches and non-matches, maybe doing [set comparisons](https://docs.python.org/3.12/library/stdtypes.html#set) of the alma and aspace dictionary keys would work?

python/requirements.txt
```diff
@@ -1 +1,3 @@
 archivessnake ~= 0.10.1
+# alma_api_client must be installed from the GitHub repository
+git+https://github.com/UCLALibrary/alma-api-client
```
akohler, line 3: Please add a line return

python/tests/test_bradley_mapping.py
```diff
@@ -0,0 +1,86 @@
+import unittest
+import sys
+import os
+import json
+
+
+# Add the config directory to the sys.path
+sys.path.append(
```
akohler, line 8: I agree this is messy, but don't offhand know the solution.  See if making this code more package-like with `__init__.py` files helps, but for now this is secondary to the suggested code restructuring elsewhere.

python/Dockerfile
akohler, line 1: Please bump the base image to `python:3.12-slim-bookworm` we're using with the Django apps now.

python/add_alma_barcodes_to_archivesspace.py
```diff
@@ -0,0 +1,120 @@
...
+    )
+
+    # update ASpace top containers with barcodes
+    for tc in matched_aspace_containers:
+        aspace_client.post(tc["uri"], json=tc)
+        logger.info(f"Added barcode to top container {tc['uri']}")
+
+    logger.info(f"Updated barcodes for {len(matched_aspace_containers)} top containers")
+
+    # if there are unmatched items or containers, write them to JSON files
```
akohler, line 105: There is some duplicated code, though refactoring is not necessarily better - I'm currently on the fence.  Maybe an convenience method which takes a `list[dict]`, a filename, and a message template, does what these do?  That may be too messy, in which case a simpler method which just takes the data and the filename would be enough - mainly for consistent file handling.  That would also apply to line 87 above, so I think 3 uses of this "write json to file" method are worthwhile.

akohler (changes requested):
@ztucker4 This new code does what it should.  My numbers and performance stats match yours, and tests pass.  I logged into my local ASpace after running it, and containers like http://localhost:8080/top_containers/141504 now include barcodes.

As we talked about, I think the config code can be refactored, both for performance and (I hope) overall simplicity, by abstracting the specific Alma and ASpace data elements needed for matching into methods returning dicts or sets of matching tuples.   The goal is a more generic `match_containers` method which delegates the details of pulling out project-specific data elements to separate methods, so for future collections we can focus just on those Alma- and ASpace-specific methods.

Please let me know if any questions.  Thanks --Andy

ztucker4:
Ready for another look! Changes:
* In `bradley.py`:
  * The main `match_containers()` function has been updated with correct type hints and a docstring describing parameters and return values. Let me know if this needs clarification or reformatting.
  * The logger is no longer included as a parameter, and is instead obtained from `globals()`.
  * Matching logic has been refactored. Two new functions, `_get_aspace_match_data()` and `_get_alma_match_data()`, create dicts with "mapping key" tuple keys and JSON object values.  Only a single loop is then used through the Alma items to compare keys and find matches.
  * Finding unmatched containers is now simplified through set comparison logic in lines 41-51.
* In `add_alma_barcodes_to_archivesspace.py`:
  * Writing JSON output files is now handled by `write_json_to_file()`, which takes the output data, a filename, and a log message string. This is used to output containers with preexisting barcodes, unmatched ASpace containers, and unmatched Alma items.
* In `test_bradley_mapping.py`:
  * By adding an `__init__.py` in the tests directory, I can now import the `config.bradley` more cleanly. However, now the tests only work when run from the parent `python` directory (`python -m unittest tests/test_bradley_mapping.py`). I'm not sure what is going on here, but it seems slightly better than having to run tests from inside the tests directory.
* Dockerfile now uses updated `python:3.12-slim-bookworm` base image.
* requirements.txt is correctly formatted.

python/add_alma_barcodes_to_archivesspace.py
```diff
@@ -0,0 +1,127 @@
...
+    while len(alma_items) < total_items:
+        current_items = alma_client.get_items(
+            bib_id, holdings_id, {"limit": 100, "offset": offset}
+        )
+        offset += 100
+        alma_items.extend(current_items.get("item"))
+    return alma_items
+
+
+def write_json_to_file(data: list[dict], filename: str, log_message: str) -> None:
```
akohler, line 46: I think it's cleaner to not have this log a message, unless the message can always be built from the other parameters.  For example, if we always wanted to log `filename` as part of the message with other parts constant, like `f"Wrote len{data} items to {filename}"`.  Since the constant text varies, that's not a good fit, and we can't reduce code similarity easily there.

Having the log written in the same method as the file write itself seems good from an atomic POV - if the file write fails, an exception will be thrown here and the log will not be written.  But that's also true if the log message is just written where it's defined.

So, let's not overload `write_json_to_file()` with `log_message`, and just log messages directly where needed instead of passing them around.  This method is still worthwhile for its main purpose.

akohler (changes requested):
@ztucker4 Excellent rewrite of the code we talked about.  I think it reads more cleanly, and will make it simpler to create different configs for future collections.

On further review, overloading `write_json_to_file()` with the `log_message` parameter does not seem like a good fit.  That's the only change I think is still needed, then I'll merge.

For the tests: they run cleanly, and running from the top-level directory seems better than having to change in and out of the `tests` directory.  This code can stay as-is, though while you're gone I think I'll experiment further with that, as I need to improve my knowledge in this area.

Thanks --Andy

ztucker4:
Requested changes made - `write_json_to_file()` now only takes in the output data and filename, and logging is handled separately.

python/config/bradley.py
```diff
@@ -0,0 +1,82 @@
...
+def _get_alma_match_data(alma_items: list) -> dict[tuple]:
+    match_data = {}
+    for item in alma_items:
+        description = item.get("item_data").get("description")
+        # split description into container type and indicator, e.g. "box.1"
+        alma_container_type = description.split(".")[0]
+        alma_indicator = description.split(".")[1]
+
+        # if indicator starts with leading zeroes, remove them
+        while alma_indicator[0] == "0":
```
akohler, line 74: I'm sorry, I thought I added this in the last review but didn't.... this should just be `alma_indicator = alma_indicator.lstrip("0")`, to remove leading zeros.

akohler (approved):
@ztucker4 Sorry, I forgot to comment on something I saw last time, for stripping leading characters.  Fix that and I promise to merge!
Thanks --Andy

python/config/bradley.py
ztucker4: Done!

PR #9, opened 2024-10-25
SYS-1692: Config to support LAUSD barcoding

ztucker4:
Relates to [SYS-1692](https://uclalibrary.atlassian.net/browse/SYS-1692)

A few changes to support barcoding the LAUSD collection. 

* This collection uses the same Description mapping logic as the Tom Bradley collection. The config file for this mapping has been renamed (`bradley.py` -> `box_description_mapping.py`) and comments updated to remove Bradley references.
  * The mapping config file now checks for duplicate item descriptions in Alma data and duplicate (container type, indicator) pairs in ASpace data - i.e. objects that would create duplicate keys in the `match_data` dicts. If duplicates are found, an error message is logged and all duplicate objects are skipped. Currently, this is only reported through the log messages. I wasn't sure if the duplicate items should be added to the output "unmatched" files or maybe output in a new set of files. 
  * The previous approach of using `globals()` to access the logger within the config file was not working. I have reverted to my initial approach of passing the logger explicitly to each function in `box_description_mapping.py` that needs to write output. Logs are now somewhat more verbose as the "matched Alma item" lines are now included.
* In the main script, `get_aspace_containers()` now deduplicates the `container_refs` obtained from the API. Also, `is_linked_to_published_record` is now checked for each top container, and any non-linked containers will be logged and skipped. (As far as I can tell, this shouldn't happen - the API should only be returning linked containers.)
* The existing Bradley tests are updated to use the new config file name.

To test barcoding the LAUSD collection locally:
```
--alma_environment sandbox \
--bib_id 9969119863606533 \
--holdings_id 22531983170006533 \
--resource_id 1415 \
--profile config.box_description_matching \
--asnake_config .archivessnake_secret_DEV.yml
```

[SYS-1692]: https://uclalibrary.atlassian.net/browse/SYS-1692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/add_alma_barcodes_to_archivesspace.py
+++ b/python/add_alma_barcodes_to_archivesspace.py
@@ -113,7 +113,7 @@ if __name__ == "__main__":
 
     # update ASpace top containers with barcodes
     for tc in matched_aspace_containers:
-        # aspace_client.post(tc["uri"], json=tc)
+        aspace_client.post(tc["uri"], json=tc)
         logger.info(f"Added barcode to top container {tc['uri']}")
 
     logger.info(f"Updated barcodes for {len(matched_aspace_containers)} top containers")
--- a/python/add_alma_barcodes_to_archivesspace.py
+++ b/python/add_alma_barcodes_to_archivesspace.py
@@ -14,13 +14,24 @@ logger = logging.get_logger("add_barcodes_to_archivesspace")
 
 def get_aspace_containers(aspace_client: ASnakeClient, resource_id: int) -> list[str]:
     url = f"/repositories/2/resources/{resource_id}/top_containers"
-    container_refs = []
-    for tc in aspace_client.get_paged(url):
-        container_refs.append(tc)
-    # this endpoint returns refs, so we need to get the full container JSON
+    container_refs = aspace_client.get(url).json()
+    # remove duplicate refs, if any
+    # deduplicate using a set of tuples of sorted items
+    container_refs_deduped = {tuple(sorted(tc.items())) for tc in container_refs}
+    # convert strings back to dicts, then set back to list
+    container_refs_deduped = [dict(tc) for tc in container_refs_deduped]
+
+    # the top containers endpoint returns refs, so we need to get the full container JSON
     containers = []
-    for tc in container_refs:
+    for tc in container_refs_deduped:
         tc_json = aspace_client.get(tc["ref"]).json()
+        # check that the container is linked to a published resource
+        if not tc_json.get("is_linked_to_published_record"):
+            logger.info(
+                f"Top container {tc_json.get('uri')} is not linked to a published resource"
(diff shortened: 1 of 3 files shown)


akohler (changes requested):
@ztucker4 A couple of changes requested, for simplification of the container de-duping.

One change which might be worth exploring: The Alma response has a lot of data, almost all of which we don't need.  Keeping just the `item_data` dictionary as `item` would allow simpler code in several sections.  I don't think we'll need any of the bib or holdings data from the response.

I ran this locally for LAUSD and it seems like something isn't getting captured.
* 2792 in Alma
* 2790 in ASpace
* 2784 matched item (alma matched aspace)
* 2784 added barcode
* 8 alma items did not match aspace

What happened to the 6 ASpace containers which apparently did not get updated?  The log doesn't show any details and `unmatched_aspace_containers.json` did not get updated.  These probably overlap with the 8 unmatched Alma items, but not being logged?

Thanks --Andy

ztucker4:
Hi @akohler, this is ready for another look. I've made the changes we discussed this morning:
* Logs are now output to `add_alma_barcodes_to_archivesspace_{timestamp}.log`. 
* `match_containers()` in the config file now returns only two elements: a list of matched top containers, and a dict containing four types of unhandled data (unmatched tcs, unmatched items, duplicate tcs, and duplicate items). 
* All output is now written to the single file `add_alma_barcodes_to_archivesspace_unhandled_{timestamp}.json`. This includes the four types of data output by `match_containers()`, plus any top containers with pre-existing barcodes.
* At the end of a run, logs now include an overview of the counts of updated and unhandled top containers and items, for easy sanity checking.

akohler (changes requested):
@ztucker4 I like the changes, which I think make the flow clearer.  I reset my local db and ran this, reviewed the log and the unhandled data, and don't see any inconsistencies.

I am requesting a few more small changes: include timestamp as part of the base filename instead of assembling it twice, and fix something in the documentation comments.  Then I'll merge.

Thanks --Andy

ztucker4:
Changes made, ready for re-review.

akohler (approved):
@ztucker4 All looks great now - merging.

akohler:
Oops, tests are failing: `docker compose run python python -m unittest tests/test_bradley_mapping.py`

ztucker4:
Tests are now fixed (two changes were needed: use the new output structure from `match_containers()`, and use only the `"item_data"` value from the full Alma response). To run: `docker compose run python python -m unittest tests/test_bradley_mapping.py`

akohler:
Confirmed tests now pass - merged.

PR #10, opened 2024-10-31
SYS-1695: Add MySQL support for top container retrieval

akohler:
Implements [SYS-1695](https://uclalibrary.atlassian.net/browse/SYS-1695).

This PR adds support for MySQL, primarily as an optional way to retrieve top container refs from database.  This is needed when a resource has more than 3000 or so top containers, as the `/repositories/{repo_id}/resources/{resource_id}/top_containers` API can time out (at least in our hosted test and production environments).

Relevant changes:
* Added `mysqlclient` python package and dependencies to the `python` docker container
* Modified `add_alma_barcodes_to_archivesspace.py` to retrieval of top container refs from database
  * Refactored `get_aspace_containers()` method to extract API retrieval of TC refs into internal method `_get_container_refs_from_api()`
  * Added internal method `_get_container_refs_from_db()`, which does the same as `_get_container_refs_from_api()` except via database query
  * Added `--use_db` command line parameter.  This defaults to `False`: if not specified, API will be used.  If command is invoked with `--use_db`, database will be used.
* Added basic test class and case to confirm that data retrieved via API and via database is identical.
  * This test assumes local database and API system, and is hard-coded to use a small resource (3002, Student Activism Materials, which currently has only 7 top containers).
  * I tested large resources as well, with 3000+ and 4000+ top containers, but these take several minutes to run.

Minor changes:
* Updated python to 3.13 for REPL improvements when exploring data
* Allowed access to production ArchivesSpace API endpoints from within the `python` docker container (assumes required SSH tunnels are in place on host)
* Added `devcontainer` support for VSCode, so we don't have to manage installation of dependencies for a virtual environment.
  * To use this feature, start VSCode from the top level directory of the project
  * VSCode should see `.devcontainer/devcontainer.json`; click "Reopen in Container" to use it.
  * This is optional, but I like it; we can look more at it at next standup.
 
Testing:
**Required**: Update `python/.archivessnake_secret_DEV.yml` to add
```
database:
  host: db
  database: archivesspace
  user: as
  password: as123
```

* Rebuild container to add mysql support: `docker compose build python`
* Run all tests: should be 5, all passing: `docker compose run python python -m unittest discover tests`
  * Or to run just the new one: `docker compose run python python -m unittest tests/test_container_retrieval.py`
* If desired, use the new `--use_db` option to barcode a collection locally.  For example, to do the Sontag collection (quick, just 370 containers or so):
```
--alma_environment sandbox \
--bib_id 9961804743606533 \
--holdings_id 22378568230006533 \
--resource_id 2361 \
--profile config.box_description_matching \
--asnake_config .archivessnake_secret_DEV.yml \
--use_db
```

[SYS-1695]: https://uclalibrary.atlassian.net/browse/SYS-1695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ztucker4 (approved):
This all looks good to me. Tests pass, and using the `--use_db` option to barcode a collection works as expected. Code changes make sense and the documentation is helpful. I also tried out the VSCode devcontainer, which is working but I'm not sure why it's useful - I'd like to talk about this more at standup. Merging now!

PR #11, opened 2024-11-08
SYS-1698: new config profile for RKO collection

ztucker4:
Implements [SYS-1698](https://uclalibrary.atlassian.net/browse/SYS-1698)

Adds a new config profile, `series_description_matching.py`, intended for barcoding the collection "RKO Radio Pictures Records" (PASC-0003).

Like other collections, the mapping here is based entirely on matching Alma descriptions with ArchivesSpace indicators and types. Unlike other collections, the descriptions and indicators contain information about Series as well as box number. Thus, extra logic is needed in the `_get_*_match_data()` functions to parse out series, indicator, and type. These three strings are then used to construct the matching key, and `match_containers()` uses these keys with no changes from `box_description_matching.py`.

The logic to parse Alma descriptions is fairly simple, as they all appear to be formatted as `"ser.X box.1234"`. ArchivesSpace data is more complex - some indicators are formatted as "123SER", while others are "SER-456". As far as I can tell, the formatting does not matter. There is only one pair of top containers in the database copy with overlapping but differently formatted indicators: [S-1468](http://localhost:8080/top_containers/44258) and [1468S](http://localhost:8080/top_containers/45450). There is only one Alma item with description `"ser.S box.1468"` (PID 23437872040006533), so I think one of the top containers may be mislabeled.

To run locally using sandbox Alma data:

```
--alma_environment sandbox \
--bib_id 9942336723606533 \
--holdings_id 22437840450006533 \
--resource_id 1273 \
--profile config.series_description_matching \
--asnake_config .archivessnake_secret_DEV.yml
```

This finds 4085 Alma items and 3679 top containers, and matches 3660 of the top containers. 17 top containers have no match, and two have duplicate keys (S-1468 and 1468S mentioned above). 425 Alma items have no match. 

[SYS-1698]: https://uclalibrary.atlassian.net/browse/SYS-1698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python/config/series_description_matching.py
```diff
@@ -0,0 +1,180 @@
...
+                    tc_indicator += char
+                else:
+                    break
+            tc_series = tc_indicator_with_series[len(tc_indicator) :]
+        # if indicator does not begin with a number, also split it into indicator and series
+        # hyphen delimited
+        # e.g. "SR-130" -> indicator "130", series "SR"
+        else:
+            tc_indicator = tc_indicator_with_series.split("-")[1]
+            tc_series = tc_indicator_with_series.split("-")[0]
```
akohler, line 99: This can be simplified somewhat, via regular expressions.  The core logic boils down to this, since there are only 2 cases to handle:
```
      if tc_indicator_with_series[0].isdigit():
          # DigitsLetter
          (tc_indicator, tc_series) = re.findall(r"(\d+)(\w+)", tc_indicator_with_series)[0]
      else:
          # Letters-Digits
          (tc_series, tc_indicator) = re.findall(r"(\w+)-(\d+)", tc_indicator_with_series)[0]
```
`re` comes from `import re`.  In both cases here, we look for 2 pieces of data:
1.  digit(s), via `\d+`
2. letter(s), via `\w+`
Each is wrapped in `{}`, to be "captured".  When multiple elements are captured like this, [re.findall()](https://docs.python.org/3/library/re.html#re.findall) returns a list of tuples.  We only want the first match, and we trust that this never returns `None` since the list indexing will fail. 

Since there's some complexity, I think it's worth wrapping this bit of code in its own function, and adding a check for `None`, at least.  Might also want to do something if multiple matches are found - I see at least one oddify in `'S-1464, S-1466 and S-1467'`.

I wrote a little stand-alone program / harness for this, to dump all the values to file for easier review.  It looks pretty clean to me but please double-check and change as needed.
```
import re
from add_alma_barcodes_to_archivesspace import _get_container_refs_from_db, get_aspace_containers
from asnake.client import ASnakeClient
from MySQLdb import connect
aspace_client = ASnakeClient(config_file=".archivessnake_secret_DEV.yml")
db_settings = aspace_client.config.get("database")
mysql_client = connect(
     host=db_settings.get("host"),
     database=db_settings.get("database"),
     user=db_settings.get("user"),
     password=db_settings.get("password"),
)

# 1273 is RKO resource
resource_id = 1273
container_refs = _get_container_refs_from_db(mysql_client, resource_id)
aspace_containers = get_aspace_containers(aspace_client, container_refs)

# DigitsLetter, or Letters-Digits, other than
# 'S-1464, S-1466 and S-1467',

with open(f"indicator_series_{resource_id}_dump.txt", "w") as f:
  for tc in aspace_containers:
      tc_indicator_with_series = tc.get("indicator")
      if tc_indicator_with_series[0].isdigit():
          # DigitsLetter
          (tc_indicator, tc_series) = re.findall(r"(\d+)(\w+)", tc_indicator_with_series)[0]
      else:
          # Letters-Digits
          (tc_series, tc_indicator) = re.findall(r"(\w+)-(\d+)", tc_indicator_with_series)[0]
      f.write(f"{tc_indicator_with_series}\t{tc_indicator}\t{tc_series}\n")
```

python/config/series_description_matching.py
```diff
@@ -0,0 +1,180 @@
...
+    Returns a dictionary with the normalized indicator, type, and series as keys, and
+    a list of items with duplicate keys.
+    """
+    match_data = {}
+    items_with_duplicate_keys = []
+    for item in alma_items:
+        description = item.get("description")
+        # split description into series and container type/indicator (space and period delimited)
+        # e.g. "ser.P box.0011" -> "P", "box", "0011"
+        alma_series = description.split(" ")[0].split(".")[1]
```
akohler, line 144: Since the Alma descriptions appear to be uniform, I think this code is fine as is.  It's clearly commented and doesn't lend itself to "cleverness" - no easy trick (that I see) to grab the 3 elements via a single operation.

akohler (changes requested):
@ztucker4 As the comments say, some refactoring to pull the now-duplicated `match_containers()` method into its own module, and to simplify the container indicator / series parsing.

I also found a change is needed in `add_alma_barcodes_to_archivesspace.py` itself, which I can't reference directly in this PR since the file did not change.  `get_aspace_containers()` should take `aspace_client` as a parameter... I can't figure out why this code actually works now, though it does.... I found this when writing the sample code for the regex change, and couldn't get container details because that method didn't know about `aspace_client`.  Simple fix, here are the two places that need changing:

```
-def get_aspace_containers(container_refs: set[str]) -> list[str]:
+def get_aspace_containers(
+    aspace_client: ASnakeClient, container_refs: set[str]
+) -> list[str]:

-    aspace_containers = get_aspace_containers(container_refs)
+    aspace_containers = get_aspace_containers(aspace_client, container_refs)
```

Thanks --Andy

ztucker4:
Ready for re-review, with more expansive changes than in the first draft:

* `match_containers()` is now included in a separate module (`config/base_match.py`). This function is now used directly in the main script, as are the two `get_match_data()` functions imported from the config file specified at runtime. Both config profiles are updated to support this.
* Within `series_description_matching.py`, regex-based extraction of indicator and series data is now done with a separate function, `parse_aspace_indicator()`. This function returns `None` for both indicator and series if `re.findall()` finds zero or multiple matches for the indicator and series. Then, `get_aspace_match_data()` checks if either indicator or series is `None` - if so, an error is logged and the TC is assigned a unique key in the matching dictionary. The TC will then not find a match in the Alma data and will be included in the output `unmatched_aspace_containers`.
* Main script now has `get_aspace_containers()` use `aspace_client` as a parameter.
* Three new tests are added for the new config profile, and existing tests are updated to work with `match_containers()` changes. To run all 8 tests: `docker compose run python python -m unittest tests/test_series_mapping.py`

PR #13, opened 2024-12-04
SYS-1713: Config file for indicator only mapping

ztucker4:
Implements [SYS-1713](https://uclalibrary.atlassian.net/browse/SYS-1713)

Adds a new config file (`indicator_only_matching.py`) intended for use with the Neutra collection. This is very similar to the previous `box_description_matching.py` file, but it relaxes the matching conditions such that two objects with the same numerical indicators will match, regardless of their types. Specifically, this should cause Alma items with descriptions of the form `box.1234` to match top container `Folder 1234` or `Box 1234`.

In addition, this PR renames the existing `box_description_matching.py` config file to `indicator_type_matching.py`. I think this is a better description of the desired mapping outcome, as nothing about this is specific to boxes.

Tests are included for the new file (see `test_indicator_only_matching.py`) and existing tests are updated to use the new `indicator_type_matching` name. A total of 14 tests should run and pass: `docker compose run --rm python python -m unittest discover`

To try the new config file with the Neutra collection, run:

```
docker compose run --rm python python add_alma_barcodes_to_archivesspace.py \
--alma_environment sandbox \
--bib_id 9924703433606533 \
--holdings_id 22441495240006533 \
--resource_id 3 \
--profile config.indicator_only_matching \
--asnake_config .archivessnake_secret_DEV.yml \
--dry_run
```

For me, this reduces the number of unmatched items in Alma and ASpace from 763 and 427 to 364 and 26, respectively. Two new ASpace containers are identified as duplicates ([Folder 2](http://localhost:8080/top_containers/161135) and [Box 2](http://localhost:8080/top_containers/408)).

[SYS-1713]: https://uclalibrary.atlassian.net/browse/SYS-1713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


akohler (approved):
@ztucker4 This looks great.  The tests all pass, and my local dry run numbers match yours.

I'm uncomfortable with the large-scale code duplication between `indicator_only_matching.py` and `indicator_type_matching.py`.  However, I'm not yet convinced it would be cleaner to try passing a new flag (`--ignore_type` maybe) which could allow code reuse by setting the `type` values to `None` - feels awkward either way.  I think the need to rewrite these config files is low, maybe zero... but if we do need to modify logic in multiple places, or start looking at a 3rd minor variant, I will want to stop for refactoring.

Merging now, thanks. --Andy

PR #14, opened 2024-12-12
SYS-1715: Refactor barcode script part 1

akohler:
Implements the first part of [SYS-1715](https://uclalibrary.atlassian.net/browse/SYS-1715):
> Main script is refactored to reduce the amount of code in the main method, improving readability and possible code reuse

This is a partial refactoring of the main script, intended mainly to pull code out of the `main()` method and to make it easier to read / understand / reason about the code.  Even that part is not yet finished: as line 379 says, there's still a lot of code in `main()` which needs to be pulled out into separate methods, and I haven't even tackled the profiles which were my original main intent.

I've stopped here for this PR to (a) break the changes into digestible (though still large) chunks, and (b) have a chance to talk about the changes and hopefully confirm this is going in a useful direction.

Changes so far:
* Code for boilerplate setup (logging, arg parsing etc.) is now in helper methods like `_get_logger()`, `_get_args()`, `_get_alma_api_key()`.  These generally have potential for reuse.
* Code for retrieving data from Alma and ASpace is split into helper methods which fetch data via API (or DB) from remote systems, and public methods which try to load cached data from file (if requested and if available), or they call the helper methods to fetch fresh data as appropriate.
* Added `--use_cache` command-line argument. If not specified, data is always fetched from remote systems.  If specified, data is loaded from files (names automatically based on data type and record id) if available, falling back to fresh data if no file(s) exist.
  * If fresh data is retrieved, it's always stored in cache files for optional future reuse.

If this PR is accepted and merged, the next step will be to finish refactoring `main()` by pulling the broad data matching code, along with the code related to `format_unhandled_data_for_printing()` and `get_run_summary_info()`, out of `main()`.  Then, the final (!) step will be to tackle the profile handling itself.

If this all somehow makes sense, great! - please merge.  We can also use standup to talk through this.

Testing: Bring up your local system.  Once ready, test with the tiny Student Activism collection:
```
docker compose run python \
python add_alma_barcodes_to_archivesspace.py \
--alma_environment sandbox \
--bib_id 9935203453606533 \
--holdings_id 22534560750006533 \
--resource_id 3002 \
--use_cache \
--dry_run \
--profile config.indicator_type_matching \
--asnake_config .archivessnake_secret_DEV.yml
```

The first run should result in a brief log file and a few json files:
```
-rw-r--r-- 1 akohler akohler   12738 Dec 11 16:02 alma_data_22534560750006533.json
[...]

PR #15, opened 2024-12-13
SYS-1715: Refactor barcode script part 2

akohler:
Implements more refactoring for [SYS-1715](https://uclalibrary.atlassian.net/browse/SYS-1715).

This PR builds on work started in #14:
* `format_unhandled_data_for_printing()` now is `print_unhandled_data()`.  Instead of building a list of strings for the `main()` method to print, it just prints the same data directly to console.
* `get_run_summary_info()` now is `print_summary_info()`. It still builds a list of strings with the same info as before, but now writes those directly to log (and to console as well, if `print_output` is True) instead of returning the list to `main()` to log & print.
* `_get_alma_api_key()` now retrieves the Alma API key from the `alma_config.alma_api_key` in the YAML config file already used for ASpace/ASnake configuration info.
  * Since the config file now contains both ASpace and Alma info, the command-line argument has changed from `--asnake_config` to `--config_file`.
  * Since the Alma environment is no longer directly relevant, the `--alma_environment` command-line argument has been removed.  If it's important, I can add it to the config file as a key, read and log it as needed... but we don't do that with the ASpace environment, so the new behavior is consistent with that.

Testing:
Add the following to `.archivessnake_secret_DEV.yml` as a top-level key:
```
alma_config:
  # Alma sandbox, during development
  alma_api_key: SANDBOX KEY GOES HERE
```

Then, with the local system up and ready, run this to do a dry run of the Student Activism collection, with output printed as well as logged as relevant:
```
docker compose run python \
--bib_id 9935203453606533 \
--holdings_id 22534560750006533 \
--resource_id 3002 \
--use_cache \
--dry_run \
--print_output \
--config_file .archivessnake_secret_DEV.yml
```

Output and logs should look the same as before.
Optionally, delete the cached `_data_*.json` files so Alma data is freshly retrieved, confirming that the Alma API key is being used.

Once this is approved and merged, I'll start refactoring the actual profile matching & related code.

[SYS-1715]: https://uclalibrary.atlassian.net/browse/SYS-1715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/add_alma_barcodes_to_archivesspace.py
+++ b/python/add_alma_barcodes_to_archivesspace.py
@@ -322,14 +322,16 @@ def print_unhandled_data(unhandled_data: dict) -> None:
         print(f"{tc}\n")
 
 
-def get_run_summary_info(
+def print_summary_info(
     alma_items: list[dict],
     aspace_containers: list[dict],
     matched_aspace_containers: list[dict],
     unhandled_data: dict,
-) -> list[str]:
+    print_output: bool,
+) -> None:
     """
-    Returns a list of strings with summary information about the run.
+    Writes summary information about the run to the log.
+    If print_output is True, also prints the info to console.
     """
     summary_info = [
         f"Total Alma items: {len(alma_items)}",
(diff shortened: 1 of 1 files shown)


ztucker4 (approved):
This all looks good. The script works as before. Moving the Alma API key to the existing config file is convenient. The updated `print` functions make sense and work to simplify the main function logic. Merging!

PR #17, opened 2025-07-02
SYS-1878: Add script to count containers for each resource

henry-r18:
Implements [SYS-1878](https://uclalibrary.atlassian.net/browse/SYS-1878)

### Acceptance criteria
- [x] The LSC barcoding project data from Airtable is extended with counts of containers for each identified resource.

### Description
This PR adds a script that accepts a CSV export of LSC's barcoding project data from Airtable (see Jira issue for link), then uses an existing function from the main barcoding script in this repo to count the number of containers related to each resource ID. The counts are then appended to the input data as a column and written to a new CSV.

The Jira ticket says to "obtain the number of ASpace records in each collection," which I interpreted to mean the number of containers related to each resource. I figured that will give us a good indication of which `Rec IDs` to prioritize. Please let me know if that's wrong.

#### Running the script
The script accepts two arguments:
1. `--file_name`: path to the CSV export of LSC's Airtable data
2. `--config_file`: path to the YAML configuration file, which includes database credentials for the target ArchivesSpace instance

With a `bash` session open in the `python` container, run `python get_container_counts.py --file_name {{PATH_TO_LSC_DATA}} --config_file {{PATH_TO_CONFIG_YAML}}`.

### Testing
I did not add any new unit tests, as the `_get_container_refs_from_db` from `add_alma_barcodes_to_archivesspace.py` already has a test. However, that test was failing, so I fixed an issue with it, which I describe inline in a separate comment.

[SYS-1878]: https://uclalibrary.atlassian.net/browse/SYS-1878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/tests/test_container_retrieval.py
+++ b/python/tests/test_container_retrieval.py
@@ -12,21 +12,11 @@ class TestContainerRetrieval(unittest.TestCase):
     def setUpClass(cls):
         # Assumes config file with local information
         cls.aspace_client = ASnakeClient(config_file=".archivessnake_secret_DEV.yml")
-        db_settings = cls.aspace_client.config.get("database")
-        cls.mysql_client = connect(
-            host=db_settings.get("host"),
-            database=db_settings.get("database"),
-            user=db_settings.get("user"),
-            password=db_settings.get("password"),
-        )
-
-    @classmethod
-    def tearDownClass(cls):
-        cls.mysql_client.close()
+        cls.db_settings = cls.aspace_client.config.get("database")
 
     def test_retrievals_are_identical(self):
         resource_id = 3002  # should have just a few top containers
         api_refs = _get_container_refs_from_api(self.aspace_client, resource_id)
-        db_refs = _get_container_refs_from_db(self.mysql_client, resource_id)
+        db_refs = _get_container_refs_from_db(self.db_settings, resource_id)
         # Don't check sizes as local data can change, but confirm the values are the same.
         self.assertEqual(api_refs, db_refs)
--- a/python/tests/test_container_retrieval.py
+++ b/python/tests/test_container_retrieval.py
@@ -1,6 +1,5 @@
 import unittest
 from asnake.client import ASnakeClient
-from MySQLdb import connect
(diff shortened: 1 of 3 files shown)


python/tests/test_container_retrieval.py
```diff
@@ -12,21 +12,11 @@ class TestContainerRetrieval(unittest.TestCase):
...
-
-    @classmethod
-    def tearDownClass(cls):
-        cls.mysql_client.close()
+        cls.db_settings = cls.aspace_client.config.get("database")
 
     def test_retrievals_are_identical(self):
         resource_id = 3002  # should have just a few top containers
         api_refs = _get_container_refs_from_api(self.aspace_client, resource_id)
-        db_refs = _get_container_refs_from_db(self.mysql_client, resource_id)
```
henry-r18, line 30: The `_get_container_refs_from_db` function takes a database configuration `dict`, not a MySQL `Connection` object as its first parameter. I removed the definitions for the MySQL client on the class, as well as the tear down method. The test should pass, provided a valid YAML config file exists.

akohler (changes requested):
@henry-r18 A few minor changes requested to code structure.  Functionally, it seems to do what's needed, though I'm surprised by the number of records which LSC has flagged for updating, but which are reported with 0 containers.  That could be due to old data I have locally, but can you double-check a couple of these?

Examples where I get 0:
* Rhea Carolyn Ackerman papers (resource 60)
* Los Angeles County tax collector property index map books, 1922-1926 (resource 3780)

Thanks --Andy

akohler (approved):
@henry-r18 Thanks for making the changes.  All looks fine now, merging. I'll ask LSC about the 0-container resources they've flagged for barcoding; if any changes to this are needed based on their response, we'll do that via new ticket.

Output for me on local system:
```
Counting containers for each resource ID...
1016 of 3476 rows were updated with counts.
Writing counts to barcoding_project_20250702_with_container_counts.csv...
```

PR #18, opened 2025-07-08
SYS-1878: Fix f-string problem under python 3.11

akohler:
Fixes a minor bug in `get_container_counts.py`: our server environment requires Python 3.11, which does not support f-strings with nested quotes of the same kind.  For example, `f"{row["id"]}"` works fine in Python 3.12 and later, but must be changed to `f'{row["id"]}'` in Python 3.11.

To avoid future problems like this, I downgraded the `python` container in this system to Python 3.11.  I also updated the development infrastructure to use our newer approach to dev containers based on docker compose.  That required changes to `README.md` as well.

### Testing

```
# Stop system, if needed
$ docker compose down

# Find aspace containers
$ docker images | grep archivesspace

# Drop the vscode-created dev container, if any
$ docker rmi -f vsc-archivesspace-toolkit-UNIQUE_HEX_STRING-uid

# Drop the old python container
$ docker rmi -f archivesspace-toolkit-python

# Clean up
$ docker system prune
```

Now run `code .` in the top-level project directory, Choose to "Reopen in container" and marvel as everything is (slowly...) rebuilt and started.

From your local shell, run tests:
`docker compose exec python python -m unittest`

or via vscode within the running terminal:
`python -m unittest`

Either should show 14 passing tests.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,31 +1,24 @@
-// For format details, see https://aka.ms/devcontainer.json. For config options, see the
-// README at: https://github.com/devcontainers/templates/tree/main/src/docker-existing-dockerfile
 {
-	"image": "archivesspace-toolkit-python:latest",
+	"name": "ASpace Python",
+	"dockerComposeFile": "../docker-compose.yml",
+	"service": "python",
+	"workspaceFolder": "/home/aspace/app",
 	"customizations": {
-		"vscode": {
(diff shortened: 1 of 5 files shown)


henry-r18 (approved):
@akohler All good for me. Within the dev container, `python --version` now shows `Python 3.11.13`. All 14 tests pass. Merging.

fyi// @ztucker4

PR #20, opened 2025-07-17
SYS-1917: Add script to remove barcodes based on log file

henry-r18:
Implements [SYS-1917](https://uclalibrary.atlassian.net/browse/SYS-1917)

### Acceptance criteria
- [x] A script is created that removes barcodes from ASpace records that have already been barcoded

### Description
This PR adds a function to `add_alma_barcodes_to_archivesspace.py` for removing barcodes from a specified collection in ArchivesSpace. This function is invoked using the new CLI flag `--undo_barcoding`. The function is interactive, prompting the user for confirmation before deleting existing barcodes from the collection identified by `resource_id`. The changes are intended as a utility for cases where barcodes are applied mistakenly and need to be rolled back, as happened recently with the Lloyd Wright collection.

I also tidied up the docstrings to include param and return annotations.

### Testing
I did not add any automated unit tests, as the function is interactive and requires user input. Please confirm manually that the function works as described and only proceeds with barcode deletion provided user confirmation.

Resource ID 3002 is a good test case, as it only includes a few top containers. To test, run the following command from the `python` container:

```bash
    --bib_id 9935203453606533 \
    --holdings_id 22534560750006533 \
    --resource_id 3002 \
    --config_file .archivessnake_secret_DEV.yml \
    --undo_barcoding \
    --use_log {PATH_TO_LOG} \
```
The log file provided to `--use_log` should include event messages in the format "Added barcode to top container..." for the relevant resource ID.

If used with only `--undo_barcoding` and not `--use_log`, the script should confirm twice before deleting all of a resource's barcodes.

The command can also be run with the `--dry_run` flag, to see counts of containers that would be effected, without taking any action. This is a redundancy on top of the confirmation interaction.

[SYS-1917]: https://uclalibrary.atlassian.net/browse/SYS-1917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python/add_alma_barcodes_to_archivesspace.py
```diff
@@ -352,7 +409,116 @@ def print_summary_info(
             print(message)
 
 
+def _get_container_refs_from_log_file(filename: str) -> set[str]:
```
henry-r18, line 412: This is where the main changes start. Everything else is docstring maintenance.

PR #22, opened 2026-03-05
DLSR-13: Update duplicate detection script to ignore backlog material

ztucker4:
Implements [DLSR-13](https://uclalibrary.atlassian.net/browse/DLSR-13)

Updates `find_duplicate_indicators.py` to address a special case noted by LSC. If a Top Container is linked to an Archival Object with "backlog material" in its title, it is not a duplicate and should be excluded.

The script now makes an additional API call for every TC with a shared indicator/type pair to check for any linked "backlog" AOs. If any are found, the TC is not considered a duplicate.

In the test ASpace database, Collection ID `828` is an example with valid, non-backlog duplicates, and `34` is an example with only backlog duplicates:

```
python find_duplicate_indicators.py --config_file .archivessnake_secret_TEST.yml --collection_id 828
python find_duplicate_indicators.py --config_file .archivessnake_secret_TEST.yml --collection_id 34
```

[DLSR-13]: https://uclalibrary.atlassian.net/browse/DLSR-13?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/find_duplicate_indicators.py
+++ b/python/find_duplicate_indicators.py
@@ -129,6 +129,27 @@ def get_locations_from_container_uri(
     return full_locations
 
 
+def get_linked_archival_objects_from_container_uri(
+    aspace_client: ASnakeClient, container_uri: str
+) -> list[str]:
+    """Given a container URI, returns a list of titles for the archival objects
+    linked to that container.
+
+    :param ASnakeClient aspace_client: An authenticated ASnakeClient instance.
+    :param str container_uri: The URI of the container to retrieve.
+    """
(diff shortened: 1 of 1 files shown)


akohler (approved):
@ztucker4 This seems correct to me.  Code and comments make sense, and I was able to run your two examples against the test system and got consistent results.

Merged, thanks --Andy

PR #23, opened 2026-05-07
DLSR-11: Parse and cleanup compound indicators

henry-r18:
Implements [DLSR-11](https://uclalibrary.atlassian.net/browse/DLSR-11)

### Summary of code logic
1. Get all top containers for a given collection.
2. Find containers with compound indicators.
3. Split each compound indicator into individual values.
4. For each value, reuse a matching existing container if found, or create a new one with copied metadata.
5. For each archival object linked to the compound container, add new instances linked to each single-box container.
6. Remove archival object links to the original compound container.
7. Delete the original compound container.

### Acceptance criteria
After running the script:
- [X] No archival objects lose container links
- [X] Each box-level container has a single, unambiguous Indicator
- [X] Container profiles and Instance types are preserved on newly created top containers
- [X] Compound containers no longer exist after processing

### Description
This PR adds a new script to cleanup top containers with compound indicators, i.e. those with comma-separated values or numeric ranges indicated by hyphens. The script identifies such top containers in a given resource/collection, parses the compound indicator into a list of discrete values, then creates new top containers or finds existing top containers using the values in the resulting list. It has a `dry_run` mode to log would-be changes without modifying anything, and it writes logs to a new `./logs` directory.

The script uses database queries for retrieving top container and archival object URIs, both to avoid potential timeout issues via the API on large collections and because there's no good way to retrieve a list of all archival objects linked to a particular top container via the API. All update actions are done via POST calls to the API, however.

Other changes include moving shared helper functions to the new `python/aspace_utils.py` module, for reuse across different scripts. The first couple commits address linter issues related to types in the existing `add_alma_barcodes_to_archivesspace.py` script and related scripts under `python/config`. I'll refactor `add_alma_barcodes_to_archivesspace.py` to use the helper functions in `python/aspace_utils.py` in a separate PR.

### Testing
I added unit tests in `python/tests/test_cleanup_compound_indicators_aspace.py` to cover the compound indicator parsing logic. There are now 16 unit tests for the toolkit, and all still pass after these changes. To run tests within the Python container: `docker compose exec python python -m unittest discover tests`

Julia from LSC [identified a few collections](https://uclalibrary.atlassian.net/browse/DLSR-11?focusedCommentId=712018) with multiple compound indicators for testing purposes. Using resource ID `1947` as an example, the following command can be run from within the Python container to see a dry run of the script:

`python cleanup_compound_indicators_aspace.py -c .archivessnake_secret_DEV.yml --resource_id 1947 --dry_run`

I ran the script in live mode against my local ArchivesSpace instance on resource `1947`. I can share my log file for that run separately for comparison.

[DLSR-11]: https://uclalibrary.atlassian.net/browse/DLSR-11?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/aspace_utils.py
+++ b/python/aspace_utils.py
@@ -89,7 +89,7 @@ def _get_ao_refs_for_top_container_from_db(
     # This adapts the query used in `_get_container_refs_from_db` to return
     # the set of archival object refs linked to the given top container.
     query = """
-        select distinct 
+        select distinct
             concat('/repositories/', r.repo_id, '/archival_objects/', ao.id) as ao_uri
         from resource r
         inner join archival_object ao on r.id = ao.root_record_id
--- a/python/aspace_utils.py
+++ b/python/aspace_utils.py
@@ -1,0 +1,111 @@
+"""
+Utility functions and helpers for working with ArchivesSpace.
+
+This module provides utility functions for interacting with ArchivesSpace
+that can be reused across multiple scripts in the toolkit.
+"""
+
+from asnake.client import ASnakeClient
+from MySQLdb import connect
+from MySQLdb.cursors import DictCursor
+
+
+def _get_container_refs_from_api(
+    aspace_client: ASnakeClient, resource_id: int
+) -> set[str]:
+    """Returns a de-duped set of _ref_ top container URIs for the given resource_id,
+    obtained via API call.
+    This API call can fail via timeout in hosted environments, when
+    more than a few thousand containers are associated with the resource.
+
+    :param ASnakeClient aspace_client: ASnakeClient instance.
+    :param int resource_id: ASpace resource ID for target collection.
+    :return: A set of container refs.
+    """
+    url = f"/repositories/2/resources/{resource_id}/top_containers"
+    container_refs = aspace_client.get(url).json()
+    # Extract the ref URIs and de-dup.
+    return set(tc["ref"] for tc in container_refs)
+
+
+def _get_container_refs_from_db(db_settings: dict, resource_id: int) -> set[str]:
+    """Returns a de-duped set of _ref_ top container URIs for the given resource_id,
+    obtained via database query.
+    This is intended as an alternative for resources with more than a few thousand
+    containers, as the API call may time out.
+
+    :param dict db_settings: A dict with DB connection details.
+    :param int resource_id: ASpace resource ID for target collection.
+    :return: A set of container refs.
+    """
+    mysql_client = connect(
+        host=db_settings.get("host"),
+        database=db_settings.get("database"),
+        user=db_settings.get("user"),
+        password=db_settings.get("password"),
+    )
+
+    query = """
+        select distinct
+            concat('/repositories/', r.repo_id, '/top_containers/', tc.id) as container_uri
+        from resource r
+        inner join archival_object ao on r.id = ao.root_record_id
+        inner join instance i on ao.id = i.archival_object_id
+        inner join sub_container sc on i.id = sc.instance_id
+        inner join top_container_link_rlshp tclr on sc.id = tclr.sub_container_id
+        inner join top_container tc on tclr.top_container_id = tc.id
+        where r.id = %s
+        and ao.publish = 1 -- true
+        and ao.suppressed = 0 -- false
+        order by container_uri
+    """
+    # Parameterized query requires tuple of values
+    cursor = mysql_client.cursor(DictCursor)
+    cursor.execute(query, (resource_id,))
+    container_refs = set(row["container_uri"] for row in cursor.fetchall())
+    cursor.close()
+    mysql_client.close()
+    return container_refs
+
+
+def _get_ao_refs_for_top_container_from_db(
+    db_settings: dict,
+    top_container_id: int,
+) -> list[str]:
+    """Return de-duped archival object refs linked to the given top container ID
+    via a database query. Filters for published and non-suppressed archival objects.
+
+    :param dict db_settings: A dict with DB connection details.
+    :param int top_container_id: ASpace top container ID.
+    :return: A list of archival object refs.
+    """
+    mysql_client = connect(
+        host=db_settings.get("host"),
+        database=db_settings.get("database"),
+        user=db_settings.get("user"),
+        password=db_settings.get("password"),
+    )
+
+    # This adapts the query used in `_get_container_refs_from_db` to return
+    # the set of archival object refs linked to the given top container.
+    query = """
+        select distinct 
+            concat('/repositories/', r.repo_id, '/archival_objects/', ao.id) as ao_uri
+        from resource r
+        inner join archival_object ao on r.id = ao.root_record_id
+        inner join instance i on ao.id = i.archival_object_id
+        inner join sub_container sc on i.id = sc.instance_id
+        inner join top_container_link_rlshp tclr on sc.id = tclr.sub_container_id
+        inner join top_container tc on tclr.top_container_id = tc.id
+        where tc.id = %s
+        and ao.publish = 1
+        and ao.suppressed = 0
+        order by ao_uri
+    """
+
+    cursor = mysql_client.cursor(DictCursor)
+    cursor.execute(query, (top_container_id,))
+    ao_refs = [row["ao_uri"] for row in cursor.fetchall()]
+    cursor.close()
+    mysql_client.close()
+    return ao_refs
--- a/python/cleanup_compound_indicators_aspace.py
+++ b/python/cleanup_compound_indicators_aspace.py
@@ -9,7 +9,6 @@ from pathlib import Path
 from asnake.client import ASnakeClient
 
 from aspace_utils import (
-    _get_container_refs_from_api,
     _get_container_refs_from_db,
     _get_ao_refs_for_top_container_from_db,
 )
@@ -35,14 +34,17 @@ def _configure_logging() -> None:
 def _get_args() -> argparse.Namespace:
     """Get command-line arguments for this program."""
     parser = argparse.ArgumentParser(
-        description="Cleanup compound box indicators in ArchivesSpace."
+        description="Cleanup compound indicators in ArchivesSpace."
     )
     parser.add_argument(
         "-c",
         "--config_file",
         type=str,
         required=True,
-        help="Path to YAML config file with ArchivesSpace credentials.",
+        help=(
+            "Path to YAML config file with ArchivesSpace credentials, "
+            "including database connection settings."
+        ),
     )
     parser.add_argument(
         "--repo_id",
@@ -59,14 +61,9 @@ def _get_args() -> argparse.Namespace:
         help="ArchivesSpace resource ID to process.",
     )
     parser.add_argument(
-        "--use_db",
-        action="store_true",
-        help="Get containers from database instead of API.",
-    )
-    parser.add_argument(
         "--dry_run",
         action="store_true",
-        help="Dry run: log would-be actions without updating ArchivesSpace.",
+        help="Log intended actions without updating ArchivesSpace.",
     )
     return parser.parse_args()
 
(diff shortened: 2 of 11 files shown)


python/aspace_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    """Returns a de-duped set of _ref_ top container URIs for the given resource_id,
+    obtained via API call.
+    This API call can fail via timeout in hosted environments, when
+    more than a few thousand containers are associated with the resource.
+
+    :param ASnakeClient aspace_client: ASnakeClient instance.
+    :param int resource_id: ASpace resource ID for target collection.
+    :return: A set of container refs.
+    """
+    url = f"/repositories/2/resources/{resource_id}/top_containers"
```
akohler: I know these mostly were copied as-is from `add_alma_barcodes_to_archivesspace.py`, so the repository id of `2` is hard-coded here.  Since these are currently used only by the new script, which take `repo_id` as a parameter (with a default of `2`), go ahead and change these utility methods to take `repo_id` as a parameter, with no hard-coding.

That also means the original script will break when refactored to import from `aspace_utils`, which is fine... it'll force that script away from hard-coded ids too.

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+)
+
+
+# Logger available globally within this module.
+# Configuration is done by _configure_logging(), which is called by main().
+# Made available globally so that tests can use the same logger with their own configuration.
+logger = logging.get_logger(Path(__file__).stem)
+
+
+def _configure_logging() -> None:
```
akohler: This is generic now, so this is a good time to move it into a utility module and import it.  That'll aid with refactoring other scripts, to make it all consistent.  Since it's not specific to ASpace, use a general utilities module instead of putting it into `aspace_utils.py`.

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+    # Normalize "&" and " and " (case-insensitive) to commas.
+    cleaned_indicator = re.sub(r"&|\band\b", ",", indicator, flags=re.IGNORECASE)
+    # Split on commas first to handle mixed cases, then expand any ranges in each part.
+    parts = [
+        part.strip() for part in cleaned_indicator.split(",") if part.strip()
+    ]  # `if part.strip()` filters out any empty segments
+
+    expanded: list[str] = []
+    for part in parts:
+        # Remove any square brackets
```
akohler: `strip()` only removes characters from start & end of string.  If `part` can have internal brackets, this won't remove them.  If that's OK, the comment needs clarification; if it's not, the code needs changing.

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+    :param dict compound_tc: The source top container record with a compound indicator.
+    :param str new_indicator: The indicator value for the new top container.
+    :return: The new top container record.
+    """
+    new_tc = copy.deepcopy(compound_tc)
+    new_tc["indicator"] = new_indicator
+    return new_tc
+
+
+def _post_top_container(
```
akohler: A better name would be `create_top_container`, or possibly `add_top_container`, to describe the desired action, not how it's done.  The doc string should change to match.

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+            return None
+    else:
+        new_uri = f"/repositories/{repo_id}/top_containers/{{NEW_TC_ID}}"
+    logger.info(
+        f"{'DRY RUN: Would create' if dry_run else 'Created'} new top container {new_uri}"
+    )
+    return new_uri
+
+
+def _post_archival_object(
```
akohler: Same as with containers, change `post` to `create` or `add`. EDIT: Reading further I see this actually updates things, not creates, so method should be `update_....`.

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+
+    :param ASnakeClient aspace_client: An authenticated ASnakeClient instance.
+    :param dict original_tc: The compound top container record being replaced.
+    :param list[str] new_tc_uris: URIs of individual top containers to link.
+    :param dict db_config: DB connection settings.
+    :param bool dry_run: If True, log the intended updates without POSTing.
+    """
+    # Parse ID from original TC URI,
+    # because it's much faster when querying the DB.
+    original_tc_uri = original_tc.get("uri", "")
```
akohler: If `original_tc.get("uri") is not trusted to exist, the next line will raise `ValueError` when it tries to act on this default empty string:
```
>>> original_tc_uri = ""
>>> original_tc_id = int(original_tc_uri.split("/")[-1])
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: ''
```
Either this exception must be handled, or the `get()` needs a default which will pass through all the subsequent code without problems. For example, changing the `get()` default to `"0"` would result in `original_tc_id` being set to `0`, which is an acceptable parameter for `_get_ao_refs_for_top_container_from_db()`.

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+
+        # Find the instance within the AO with ref to the original TC.
+        # The path of the ref is instance > sub_container > top_container > ref.
+        original_instances = archival_object.get("instances", [])
+        new_instances = []
+        instance_type = ""
+        for instance in original_instances:
+            sub_container = instance.get("sub_container", {})
+            top_container = sub_container.get("top_container", {})
+            if top_container.get("ref") == original_tc_uri:
```
akohler: Other `get()` references have defaults; this one does not, and will raise an exception if that key is not found.

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+                {
+                    "instance_type": instance_type,
+                    "sub_container": {"top_container": {"ref": new_tc_uri}},
+                }
+            )
+            logger.info(
+                f"Relinked archival object {ao_ref} to top container {new_tc_uri}"
+            )
+
+        # If the instances have changed, POST the updated AO to ArchivesSpace,
```
akohler: This `POST` reference will need changing when the method changes.  The action / effect is enough: "..., update the AO in ArchivesSpace...."

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+        # or log the intended update if dry_run.
+        if original_instances != new_instances:
+            archival_object["instances"] = new_instances
+            _post_archival_object(aspace_client, ao_ref, archival_object, dry_run)
+
+
+def _delete_top_container(
+    aspace_client: ASnakeClient, tc_uri: str, dry_run: bool
+) -> None:
+    """Delete the given top container URI.
```
akohler: This deletes the top container itself, not its URI.

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+        for indicator in individual_indicators:
+            # Check if the indicator already exists in the collection,
+            # and if so, add it to the list of individual URIs.
+            if indicator in existing_indicators:
+                existing_tcs = tcs_by_indicator[indicator]
+                # Log any cases where there is more than 1 TC with the same indicator
+                if len(existing_tcs) > 1:
+                    logger.warning(
+                        f"Found more than one top container with indicator '{indicator}' "
+                        f"in resource {resource_id}: {[tc.get('uri') for tc in existing_tcs]}. "
```
akohler: `tc.get('uri')` has no default and may be unsafe.

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+        except Exception as err:
+            logger.error(f"Error deleting top container {tc_uri}: {err}. Skipping.")
+            return None
+    logger.info(
+        f"{'DRY RUN: Would delete' if dry_run else 'Deleted'} "
+        f"compound top container {tc_uri}"
+    )
+
+
+def _process_resource(
```
akohler: This method does a lot, but if the doc string is accurate, the method does something more specific than the vague "process" name suggests.

Longer method names are OK.  `clean_up_box_indicators()` or something similar is a much clearer indicator (ha!) of what the method does.

python/cleanup_compound_indicators_aspace.py
```diff
@@ -0,0 +1,480 @@
...
+        # Use list of reused or new top containers to relink archival objects
+        # then delete the original compound top container.
+        _relink_archival_objects(
+            aspace_client, compound_tc, individual_uris, db_config, dry_run
+        )
+        _delete_top_container(aspace_client, compound_uri, dry_run)
+
+
+def main() -> None:
+    """Cleanup top containers in ArchivesSpace that have a compound box indicator,
```
akohler: This block of documentation is helpful, but I found it a little confusing.  `resource_id` is a required argument for the program, but the steps listed below don't mention resources at all: step 1 refers to a collection. Is that the same as a resource?

The summary in the first 2 lines doesn't refer to a resource/collection, so at first I thought that the `main()` method would iterate over everything.  Since it's focused on one thing, the summary should make that immediately clear.

akohler (changes requested):
@henry-r18 This is very good.  The code is well organized overall, with a clear general flow and appropriate utility methods.  I especially appreciate that you cleaned up type hints and some other details in existing code, improving related parts of the code base while creating this new scripts.

I've requested a lot of changes, but most of them are small: making some method names and comments a little clearer and reorganizing the code a little more.  I appreciate that you were careful with dictionary `get()` accesses, and noticed a couple which lack a default but are not wrapped in `try...except...` so could crash if data is not as you assume.

I did not try running the program tonight.  I may try tomorrow, but would prefer that @ztucker4 review that as part of the application-specific logic.  I focused mostly on the program flow and safety.

Thanks for your hard work on this --Andy

python/cleanup_compound_indicators_aspace.py
henry-r18: @akohler This function does use the `asnake.logging` module, which seems to combine some aspects of the standard `logging` module with `structlog`--see source code [here](https://github.com/archivesspace-labs/ArchivesSnake/blob/master/asnake/logging/__init__.py).

Would you rather we switch over to using the standard `logging` module, with a `configure_logging` function shared in a generic `utils.py` module for the toolkit? Or is `asnake.logging` generic enough to share with future scripts that may not directly use the `asnake` client?

The main difference I see between standard `logging` and `asnake.logging` is that the later uses JSONL by default, whereas the former uses the format we've been using in other projects.

akohler: @henry-r18 I remember `asnake` using `structlog` or something like that.  I have no real preference, and don't know how hard it would be to turn off / override what `asnake` does.
Either way, `_configure_logging()` in this script seems to be generic and suitable for sharing among our scripts within this project.  I expect most of those scripts will use the `asnake` client to some degree.

akohler (approved):
@henry-r18 These changes all look perfect to me.  Starting a formal package for the shared utilities is fine.  The code and comment updates are clear and you made a couple of extra improvements I had missed.

@ztucker4 Over to you for final review and if you have no concerns, please go ahead and merge and let Julia know this is ready.  I do recommend that she use the test instance at first.  Connecting to the database may be a challenge for her, with the extra tweaks to tunnel to the database, but you can help her with that if needed.

Thanks --Andy

ztucker4 (approved):
This all looks good to me. I've tested several individual cases and a partial run of the complete script in my local environment. The script implements the requirements correctly - "compound" TCs are split apart while maintaining AO links of the correct `type`. I'll let Julia know that this is ready and suggest that we try it in the test environment. Thanks!

PR #24, opened 2026-05-15
DLSR-305: Report container mismatches b/w ASpace and Alma

henry-r18:
Implements [DLSR-305](https://uclalibrary.atlassian.net/browse/DLSR-305)

### Acceptance criteria
- [X] All Alma items for a given collection are evaluated against the ASpace top container list
- [X] Any Alma item whose box identifier does not match an ASpace top container is captured in the output report
- [X] No data is modified in Alma or ArchivesSpace
- [X] The report is produced as a CSV file
- [X] The script completes without error when a collection has zero discrepancies, producing an empty report or a summary confirmation

### Description
This PR adds `python/find_missing_containers_aspace.py`, a new script that compares top container type-indicator values from ArchivesSpace to corresponding item descriptions in Alma, then outputs a report of Alma items that have no matching top container in ArchivesSpace. Almost all the main business logic for the script is reused from `config.base_match` and `config.indicator_type_matching`, which we used in the barcoding script `add_alma_barcodes_to_archivesspace.py`.

The basic shape of the `get_alma_items_from_alma()` function was copied from `add_alma_barcodes_to_archivesspace.py` to a new utility module `utils/alma_utils.py`. However, due to changes to [alma-api-client](https://github.com/UCLALibrary/alma-api-client), this function had to be refactored a bit from how it's implemented in `add_alma_barcodes_to_archivesspace.py`.

I chose not to use a log file or log stream in this script, opting instead for simple print statements, since the output report provides the target information. I also chose to default to the database when retrieving ASpace top container refs, both to side-step the API timeout issue on larger collections and to keep usage consistent with `cleanup_compound_indicators_aspace.py` for LSC staff.

### Testing
No new unit tests are added. All 16 existing tests still pass.

To test manually, I used the following LSC collection:
| Identifier | Title | ASpace resource ID | Alma MMS ID | Alma holdings ID |
|--------|--------|--------|--------|--------|
| LSC.1550 | Leo Kuper papers | 692 | 9919730453606533 | 22529139550006533 |

The following command should yield a CSV report showing 1 mismatched box:
`python find_missing_containers_aspace.py -c .archivessnake_secret_DEV.yml -r 692 --bib_id 9919730453606533 --holdings_id 22529139550006533`

**Expected output**
```
ASpace Resource ID,Alma Bib ID,Alma Item Barcode,Alma Box Identifier,ASpace Match Found,Notes
692,9919730453606533,FF0000132035,box.51,No,No matching ASpace top container found — requires review
```

[DLSR-305]: https://uclalibrary.atlassian.net/browse/DLSR-305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/find_missing_containers_aspace.py
+++ b/python/find_missing_containers_aspace.py
@@ -1,0 +1,216 @@
+import argparse
+import csv
+
+from alma_api_client import AlmaAPIClient
+from asnake.client import ASnakeClient
+from pathlib import Path
+
+from utils import load_config
+from utils.alma_utils import get_alma_items_from_alma
+from utils.aspace_utils import get_container_refs_from_db
+
+from config.base_match import match_containers
+from config.indicator_type_matching import get_aspace_match_data, get_alma_match_data
+
+
+def _get_args() -> argparse.Namespace:
+    """Get command-line arguments for this program."""
+    parser = argparse.ArgumentParser(
+        description=(
+            "Using ArchivesSpace and Alma IDs for the same LSC collection, "
+            "list Alma items whose box identifier has no matching top container in ArchivesSpace."
+        )
+    )
+    parser.add_argument(
+        "-c",
+        "--config_file",
+        type=str,
+        required=True,
+        help=(
+            "Path to YAML config file with ArchivesSpace credentials, "
+            "including database connection settings and Alma API key."
+        ),
+    )
+    parser.add_argument(
+        "--repo_id",
+        type=int,
+        required=False,
+        default=2,
+        help="ArchivesSpace repository ID. Defaults to 2.",
+    )
+    parser.add_argument(
+        "-r",
+        "--resource_id",
+        type=int,
+        required=True,
+        help="ArchivesSpace resource ID (i.e. collection) to process.",
+    )
+    parser.add_argument(
+        "--bib_id",
+        type=str,
+        required=True,
+        help="Alma bib MMS ID for the same collection.",
+    )
+    parser.add_argument(
+        "--holdings_id",
+        type=str,
+        required=True,
+        help="Alma holdings MMS ID for the same collection.",
+    )
+    parser.add_argument(
+        "-o",
+        "--output_path",
+        type=str,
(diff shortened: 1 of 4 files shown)


python/find_missing_containers_aspace.py
```diff
@@ -0,0 +1,228 @@
...
+def _write_report_csv(
+    output_path: Path,
+    rows: list[dict],
+) -> None:
+    """Write the CSV report to the given output path.
+
+    :param Path output_path: Path to write the CSV report.
+    :param list[dict] rows: A list of CSV row dictionaries.
+    """
+    fieldnames = [
```
akohler: Instead of hard-coding these field names, get them from `rows`:
```
fieldnames = list(rows[0].keys())
```
Since each row of data is created consistently in `_prepare_report_rows()`, and dictionaries [preserve insertion order](https://docs.python.org/3/library/stdtypes.html#dict) as of Python 3.7, this should be safe: each `dict` should have the same keys in the same order.  If they weren't created in a consistent way, this might not be safe... but then we'd probably want to use `csv.Writer` and manage everything ourselves, instead of `DictWriter`.

akohler (changes requested):
@henry-r18 Everything works as expected, the code is clear.  I'm glad existing code could be reused for the matching logic, and nothing new requires tests in my opinion.  I agree with just using print statements for the minimal non-report output needed.  I ran it with your example and got the same results.

I am requesting one minor change, to avoid hard-coding the field names in `_write_report_csv()`.  I see other scripts in this collection also did this, something I should have caught before.  If/when those are touched for refactoring or other changes, that can be fixed in those.  Making this more generic should also allow it to be a utility method instead of included in each script.

Once that small change is made, I'll merge.

Thanks --Andy

akohler (approved):
@henry-r18 Refactor looks great, merging now.

PR #25, opened 2026-05-20
DLSR-351: Refactor scripts to use shared utilities

henry-r18:
Implements [DLSR-351](https://uclalibrary.atlassian.net/browse/DLSR-351)

### Acceptance criteria
- [X] Scripts are refactored to use shared code within the local `utils` package

### Description
This PR refactors the scripts under the `python` directory to use shared code that was moved into `python/utils` in #23 and #24. Much of the refactoring has to do with logging, loading config, or writing output to CSV files. Other changes include doc-string updates and general maintenance to match our latest code style.

### Testing
All 16 unit tests in the project still pass: `python -m unittest` from within the `python` container. 

[DLSR-351]: https://uclalibrary.atlassian.net/browse/DLSR-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/add_alma_barcodes_to_archivesspace.py
+++ b/python/add_alma_barcodes_to_archivesspace.py
@@ -396,7 +396,7 @@ def main() -> None:
 
     # For convenience while debugging, print log name without full container path.
     # Also used in names of some output files.
-    logging_filename_base = Path(logging.handler.baseFilename).stem
+    logging_filename_base = Path(__file__).stem
     print(f"Logging to {logging_filename_base}.log")
     configure_logging(log_filename_stem=logging_filename_base)
 
--- a/python/add_alma_barcodes_to_archivesspace.py
+++ b/python/add_alma_barcodes_to_archivesspace.py
@@ -397,7 +397,7 @@ def main() -> None:
     # For convenience while debugging, print log name without full container path.
     # Also used in names of some output files.
     logging_filename_base = Path(logging.handler.baseFilename).stem
-    print(f"Logging to logs/{logging_filename_base}.log")
+    print(f"Logging to {logging_filename_base}.log")
     configure_logging(log_filename_stem=logging_filename_base)
 
     args: argparse.Namespace = _get_args()
--- a/python/add_alma_barcodes_to_archivesspace.py
+++ b/python/add_alma_barcodes_to_archivesspace.py
@@ -8,7 +8,7 @@ from asnake.client import ASnakeClient
 import asnake.logging as logging
 
 from config.base_match import match_containers
-from utils import configure_logging, load_config
+from utils import configure_logging, load_config, read_from_cache, write_to_cache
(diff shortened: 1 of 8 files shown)


python/utils/generic_utils.py
akohler: This needs updating, as the module is broader than "logging setup helpers" now.

python/utils/generic_utils.py
```diff
@@ -48,3 +49,38 @@ def write_dicts_to_csv(
...
         writer.writerows(rows)
+
+
+def read_from_cache(filename: str) -> list[dict] | None:
+    """Reads data from the given file and returns it.
+    Data is expected to be a list of dictionaries,
+    but this method does not enforce that.
+
+    :param str filename: Filename of cache file.
+    :return: A list of Alma items, or None if the cache file does not exist.
```
akohler: When this was `_get_cached_data_from_file()` within `add_alma_barcodes_to_archivesspace.py`, its Alma-specific purpose was clearer.  Now that it's pulled out as a more generic method, the mismatch among method name / overall doc-string / return comment is stronger.

This method currently is only used in the alma barcode script, but could be useful for non-Alma data (and I see now that script also can store ASpace data in a cache file).  Make the return comment generic.  I think the method name is OK, the return type helps.  The general comment about not enforcing the return data type... I think we can leave this as is for now, and refactor to be stricter later if / when needed.

python/add_alma_barcodes_to_archivesspace.py
```diff
@@ -529,11 +397,15 @@ def main() -> None:
     # For convenience while debugging, print log name without full container path.
     # Also used in names of some output files.
     logging_filename_base = Path(logging.handler.baseFilename).stem
-    print(f"Logging to {logging_filename_base}.log")
+    print(f"Logging to logs/{logging_filename_base}.log")
```
akohler: I'd remove "logs/" from this.  `configure_logging()` is hard-coded to put everything in `logs/` now, and I'd rather not have a hard-coded path here in this message too.  The other scripts don't have a "Logging to" message, which is fine, but when they do, let's keep it general.

akohler (changes requested):
@henry-r18 Most of this looks great, and I'm happy to see common code pulled into utility modules.  I noted a few small inconsistencies, nothing major.

Otherwise, beyond the unit tests passing, can you please confirm that the scripts themselves still function as expected?  Dry runs should be plenty.  For example, I see changes in how the ASnake client is initialized, so being certain all still works as appropriate is important.

Thanks --Andy

akohler (approved):
@henry-r18 Thank you for all of the improvements.  I think this is all set now - merging.  Thanks --Andy

PR #26, opened 2026-06-02
DLSR-260: Add script to merge duplicate top containers

henry-r18:
Implements [DLSR-260](https://uclalibrary.atlassian.net/browse/DLSR-260)

### Acceptance criteria
- [x] No archival objects lose container links as a result of the merge
- [X] Each collection contains at most one top container per unique container type + indicator combination after processing
- [x] Container profiles, barcodes, and Instance types are preserved on the canonical container
- [X] Duplicate container records no longer exist after processing
- [X] A log or summary is produced identifying which containers were merged with any errors documented

### Description
This PR adds `merge_duplicate_containers_aspace.py`, which finds and merges duplicate top containers within a single ASpace collection. After fetching top container records, the script groups duplicates using a `(type, indicator)` tuple as group keys. From these groups, canonical top containers are identified per LSC specs: either the one with the most linked archival objects, or the oldest `create_time` in the case of ties.

One thing this script does differently than `cleanup_compound_indicators_aspace.py` is it merges duplicates using the ArchivesSpace `/merge_requests/top_container` API endpoint, rather than manual AO relinking. I had overlooked this endpoint in earlier review of the ASpace docs (which to be fair is not the clearest...); relevant documentation is [here](https://archivesspace.github.io/archivesspace/api/?python#carry-out-a-merge-request-against-top-container-records). From manual testing using some scratch scripts, this endpoint appears to do everything LSC requested when consolidating top container records: identify target top container and sources, re-link all related archival objects from sources to target, then delete the sources.

The script has a dry-run mode that outputs slightly different information in a more human-readable format than the normal mode. This is intended to make manual review of the logs easier on LSC. Please let me know if you have thoughts on how to improve the clarity of the log messages or summary.

### Testing
Unit tests are added for the canonical selection logic, location-data warnings, and accession-keyword detection. There are now 22 tests for the toolkit.

As for manual testing, resource `3646` is a good illustration. There should be 54 containers before merging 13 duplicate groups, and 16 top containers after: `python merge_duplicate_containers_aspace.py -c .archivessnake_secret_DEV.yml -r 3646 -d` for dry run; `-d` can be removed for normal mode.

#### Demo gist
I created this (kind of messy) [Gist](https://gist.github.com/henry-r18/c4697f484c08147f2613cb58ea52ed0f) to demonstrate that the function that uses the `merge_requests/top_container` endpoint behaves as desired by LSC. The Gist will run "live", meaning it will actually merge the resource ID provided as a illustration in your local ASpace instance, so be warned. I picked that collection (i.e. resource ID) because it has several duplicate groups and gives a good demonstration that the endpoint manages relinked AOs from duplicates to the canonical TC and then deletes the duplicates.

I can provide a list of other good collections to test, if desired.

[DLSR-260]: https://uclalibrary.atlassian.net/browse/DLSR-260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/merge_duplicate_containers_aspace.py
+++ b/python/merge_duplicate_containers_aspace.py
@@ -129,7 +129,8 @@ def _has_location_data(tcs: list[dict]) -> bool:
         locations = tc.get("container_locations", [])
         if locations:
             logger.warning(
-                f"Top container {tc.get('uri')} has location data: {locations}"
+                f"Top container {tc.get('uri')} has location data: "
+                f"{[location.get('ref') for location in locations]}"
             )
             return True
     return False
--- a/python/merge_duplicate_containers_aspace.py
+++ b/python/merge_duplicate_containers_aspace.py
@@ -110,11 +110,12 @@ def _resolve_ao_refs_for_tcs(
     return tcs
 
 
-def _check_for_location_data(tcs: list[dict]) -> None:
+def _has_location_data(tcs: list[dict]) -> bool:
     """Check for location data on a list of top container records
     representing a duplicate group, flagging the group for review if found.
 
     :param list[dict] tcs: List of top container records representing a duplicate group.
+    :return: True if any top container has location data, False otherwise.
     """
     for tc in tcs:
         locations = tc.get("container_locations", [])
@@ -122,6 +123,8 @@ def _check_for_location_data(tcs: list[dict]) -> None:
             logger.warning(
                 f"Top container {tc.get('uri')} has location data: {locations}"
             )
+            return True
+    return False
 
 
 def _has_recent_accession_keywords(tcs: list[dict]) -> bool:
@@ -223,6 +226,40 @@ def _merge_top_containers(
     return True
 
 
+def _print_summary(summary: dict, dry_run: bool) -> None:
+    """Add summary info to the log and print a friendly message to the console.
+
+    :param dict summary: Dict containing summary info for the run.
+    :param bool dry_run: If True, print a dry run report.
+    """
+    if dry_run:
+        lines = [
+            f"{'*' * 5} DRY RUN SUMMARY {'*' * 5}",
+            f"Total duplicate groups: {summary['Total duplicate groups']}",
+            f"Groups with location data: {summary['Groups with location data']}",
+            (
+                f"Groups with recent accession keywords:"
+                f" {summary['Groups with recent accession keywords']}"
+            ),
+        ]
+    else:
+        lines = [
+            f"{'*' * 5} SUMMARY {'*' * 5}",
+            f"Total duplicate groups: {summary['Total duplicate groups']}",
+            f"Groups with location data: {summary['Groups with location data']}",
+            (
+                f"Groups with recent accession keywords:"
+                f" {summary['Groups with recent accession keywords']}"
+            ),
+            f"Successful merges: {summary['Successful merges']}",
+            f"Failed merges: {summary['Failed merges']}",
+        ]
+    lines.append(f"{'*' * len(lines[0])}")
+    for line in lines:
+        print(line)
+        logger.info(line)
+
+
 def _process_duplicates_in_collection(
     aspace_client: ASnakeClient,
     db_config: dict,
@@ -255,9 +292,10 @@ def _process_duplicates_in_collection(
 
     summary = {
         "Total duplicate groups": len(duplicate_groups),
+        "Groups with location data": 0,
+        "Groups with recent accession keywords": 0,
         "Successful merges": 0,
         "Failed merges": 0,
-        "Groups requiring manual review": 0,
     }
     for type, indicator, tcs in duplicate_groups:
         logger.info(
@@ -268,8 +306,9 @@ def _process_duplicates_in_collection(
         # Check for any location data in the duplicate group.
         # Per ticket, this should not stop processing,
         # but should be logged for visibility in dry run mode.
-        if dry_run:
-            _check_for_location_data(tcs)
+        if dry_run and _has_location_data(tcs):
+            summary["Groups with location data"] += 1
+            continue
 
         # Resolve AO refs to their full dictionaries
         # to make it easier to check AO titles for recent accession keywords
@@ -278,7 +317,7 @@ def _process_duplicates_in_collection(
         # Check for recent accession keywords in the titles of related archival objects
         # in the duplicate group, and stop processing the group if any are found.
         if _has_recent_accession_keywords(tcs):
-            summary["Groups requiring manual review"] += 1
+            summary["Groups with recent accession keywords"] += 1
             continue
 
         canonical_tc, duplicate_tcs = _determine_canonical_tc(tcs)
@@ -297,9 +336,7 @@ def _process_duplicates_in_collection(
             continue
         summary["Successful merges"] += 1
 
-    logger.info("Finished merging duplicate top containers")
-    for key, value in summary.items():
-        logger.info(f"{key}: {value}")
+    _print_summary(summary, dry_run)
 
 
 def main() -> None:
--- a/python/merge_duplicate_containers_aspace.py
+++ b/python/merge_duplicate_containers_aspace.py
@@ -280,15 +280,28 @@ def _process_duplicates_in_collection(
         5. Delete the duplicate top container(s).
     """
     tcs_by_indicator = _get_tcs_by_indicator(aspace_client, db_config, resource_id)
-    # Find TC records that have duplicate (type, indicator) keys
+    # Group TC records by (type, indicator) keys
     duplicate_groups: list[tuple[str, str, list[dict]]] = [
         (type, indicator, tcs)
         for (type, indicator), tcs in tcs_by_indicator.items()
         if len(tcs) > 1
     ]
+    # If no duplicate groups are found, log a note and return
     if not duplicate_groups:
         logger.info(f"No duplicate top containers found for Resource ID {resource_id}.")
         return
+
+    # Now sort the duplicate groups alphabetically by type and numerically by indicator,
+    # to make review of the logs easier.
+    duplicate_groups = sorted(
+        duplicate_groups,
+        key=lambda group: (
+            group[0],  # sort alphabetically by type
+            (
+                int(group[1]) if group[1].isdigit() else group[1]
+            ),  # then numerically by indicator, if possible
+        ),
+    )
 
     summary = {
         "Total duplicate groups": len(duplicate_groups),
--- a/python/merge_duplicate_containers_aspace.py
+++ b/python/merge_duplicate_containers_aspace.py
@@ -2,6 +2,7 @@ import argparse
 
 from asnake import logging
 from asnake.client import ASnakeClient
+from asnake.jsonmodel import JM
 from collections import defaultdict
 from pathlib import Path
 
@@ -28,6 +29,13 @@ def _get_args() -> argparse.Namespace:
         type=str,
         required=True,
         help="Path to YAML config file with ArchivesSpace credentials.",
+    )
+    parser.add_argument(
+        "--repo_id",
+        type=int,
+        required=False,
+        default=2,
+        help="ArchivesSpace repository ID to target. Defaults to 2.",
     )
     parser.add_argument(
         "-r",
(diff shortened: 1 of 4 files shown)


python/merge_duplicate_containers_aspace.py
```diff
@@ -0,0 +1,377 @@
...
+    parser.add_argument(
+        "-d",
+        "--dry_run",
+        action="store_true",
+        help="Run in dry run mode, without making any changes to ArchivesSpace.",
+    )
+    return parser.parse_args()
+
+
+def _get_tcs_by_indicator(
```
akohler: I think this method needs a better name.  I expected it to use an `indicator` parameter to filter the containers it returned.  Instead, it returns all of a resource's containers, grouped by type and indicator.

`cleanup_compound_indicators_aspace._get_all_top_containers()` also returns (in part) something called `tcs_by_indicator`, which really is grouped by indicator.  

I suggest `_get_tcs_grouped_by_type_and_indicator()` here, or something similar if you can shorten it with clarity.  Variable names should be changed here and in calling code to reflect this.

python/merge_duplicate_containers_aspace.py
```diff
@@ -0,0 +1,377 @@
...
+    aspace_client: ASnakeClient,
+    db_config: dict,
+    resource_id: int,
+) -> defaultdict[tuple[str, str], list[dict]]:
+    """Get all top containers in the collection grouped by (type, indicator).
```
akohler: `collection` here should be `resource`. I know they are synonyms (or we treat them as such :) ) but future maintainers may not.

python/merge_duplicate_containers_aspace.py
```diff
@@ -0,0 +1,377 @@
...
+            except Exception as err:
+                logger.error(
+                    f"Error fetching archival object {ao_ref}: {err}. Skipping."
+                )
+                continue
+    return tcs
+
+
+def _has_location_data(tcs: list[dict]) -> bool:
+    """Check for location data on a list of top container records
```
akohler: I think the "duplicate group" comments here aren't needed.  This implementation doesn't know the `tcs` list represents duplicates in the main program; it just checks to see if a data value exists or not.  Removing the "duplicate group" parts of the comments makes this method stand alone, allowing reuse in other programs if needed.

python/merge_duplicate_containers_aspace.py
```diff
@@ -0,0 +1,377 @@
...
+        if locations:
+            logger.warning(
+                f"Top container {tc.get('uri')} has location data: {locations}"
+            )
+            return True
+    return False
+
+
+def _has_recent_accession_keywords(tcs: list[dict]) -> bool:
+    """Within a top container duplicate group, check for recent accession keywords
```
akohler: Similar to `_has_location_data()`, I think the mention of duplicate groups here is not needed.  I see it is used in the log warning message; I think that sentence could become "Manual review required" or something similarly generic.  If the caller needs the extra context of `tcs` being a duplicate group, I think the caller could log an extra message if this method returns `True`.

python/merge_duplicate_containers_aspace.py
```diff
@@ -0,0 +1,377 @@
...
+    :return: A tuple of the canonical top container and the remaining duplicate TCs.
+    """
+    # Had help from LLM for this concise implementation.
+    # Selects the minimum value of the tuple returned by the lambda function,
+    # which is the TC with the most related archival objects (i.e. smallest negative value)
+    # or the oldest `create_time` if there are ties in the AO counts.
+    canonical = min(
+        tcs,
+        key=lambda tc: (
+            -len(tc["_related_aos_temp"]),
```
akohler: I know `_resolve_aos_for_tcs()` added this key so it should be guaranteed to exist, but if doesn't, this crashes.  For consistency and safety, use `.get()` with a default empty list.

python/merge_duplicate_containers_aspace.py
```diff
@@ -0,0 +1,377 @@
...
+    """Determine the canonical top container from a list of top container records.
+
+    Uses the record with the most related archival objects,
+    or the oldest creation time if there are ties.
+
+    :param list[dict] tcs: List of top container records.
+    :return: A tuple of the canonical top container and the remaining duplicate TCs.
+    """
+    # Had help from LLM for this concise implementation.
+    # Selects the minimum value of the tuple returned by the lambda function,
```
akohler, line 167: I understand the code (and appreciate the disclaimer), but I think this uses a double-negative approach which opposes the intent.  We want the largest value (the TC with the most AOs), so instead of `min()` with artificially negative numbers, use `max()` with the numbers themselves.
The `create_time` would then need to be an impossibly early value, like `"0000-01-01T00:00:00Z"` - it won't be used except for this sorting, so it probably doesn't need to be a valid `datetime`.

python/merge_duplicate_containers_aspace.py
```diff
@@ -0,0 +1,377 @@
...
+
+def _print_summary(summary: dict, dry_run: bool) -> None:
+    """Add summary info to the log and print a friendly message to the console.
+
+    :param dict summary: Dict containing summary info for the run.
+    :param bool dry_run: If True, print a dry run report.
+    """
+    if dry_run:
+        lines = [
+            f"{'*' * 5} DRY RUN SUMMARY {'*' * 5}",
```
akohler: These first lines of output are identical for dry run and production, other than the first "SUMMARY" message.  Make a `summary` variable for "SUMMARY" or "DRY RUN SUMMARY", pull the identical lines out of the conditional, and just append the final 2 Successful/Failed lines if it's production mode.

python/merge_duplicate_containers_aspace.py
```diff
@@ -0,0 +1,377 @@
...
+
+def _process_duplicates_in_collection(
+    aspace_client: ASnakeClient,
+    db_config: dict,
+    repo_id: int,
+    resource_id: int,
+    dry_run: bool,
+) -> None:
+    """Merge duplicate top containers in ArchivesSpace,
+    for a given collection (identified by `resource_id`).
```
akohler, line 308: 👍 _resource / collection synonyms_ :)

python/merge_duplicate_containers_aspace.py
```diff
@@ -0,0 +1,377 @@
...
+    duplicate_groups = sorted(
+        duplicate_groups,
+        key=lambda group: (
+            group[0],  # sort alphabetically by type
+            (
+                int(group[1]) if group[1].isdigit() else group[1]
+            ),  # then numerically by indicator, if possible
+        ),
+    )
+
```
akohler: This is a long method, which I think can be split here.  The first part is just getting `duplicate_groups`, so I would pull all of that into a separate method and call that to get `duplicate_groups`.  Then the second part of this method can remain as-is.

python/merge_duplicate_containers_aspace.py
```diff
@@ -0,0 +1,377 @@
...
+        logger.info(
+            f"Found {len(tcs)} top containers "
+            f"with type '{type}' and indicator '{indicator}'"
+        )
+
+        # Check for any location data in the duplicate group.
+        # Per ticket, this should not stop processing,
+        # but should be logged for visibility in dry run mode.
+        if dry_run and _has_location_data(tcs):
+            summary["Groups with location data"] += 1
```
akohler, line 342: This is the only place this summary value is incremented - during dry runs only - but `_print_summary()` prints it for both dry runs and production runs.  That means a production run will always claim that none of the groups have location data - it will always say `Groups with location data: 0`.  If a user failed to catch these via a dry run, the production log would not show anything about these.  I read Julia's ticket update and I'm not sure she explicitly considered that; might be worth double-checking with her.

akohler (changes requested):
@henry-r18 Overall this is excellent.  I appreciate the approach to logging (with minor caveats I'll note below) and it's interesting to see the `asnake.jsonmodel` usage here.  Tests make sense, cover the cases I can think of, and all pass.

I ran your example from the PR, in dry run mode only, and got useful output.  I haven't run your gist locally, but may do that in the morning.

I did request several changes, but almost all are small and are intended to make the code and/or comments a little clearer.  Nothing was too confusing, just rough spots which caught my mind while reading.

As for the log, 2 things should be considered:
1. The console messages should include the log file name, to make it easier to find the details after running.  That was a recent change to `add_alma_barcodes_to_archivesspace.py` and I think is worth making consistent with all of these scripts.
2. The timestamps in the log file all are in UTC: `2026-06-03T02:31:15.234130Z`.  I never noticed this before, but assume it's always been the case.  i see this is the [default set in the logging client](https://github.com/archivesspace-labs/ArchivesSnake/blob/master/asnake/logging/__init__.py#L97).  Given the use of `structlog`, I can understand that if we want to programmatically act on log messages, that's probably what ASpace expects... but even I don't automatically decode UTC mentally :) .  I suggest changing this so the logs use the timezone from the container (which gets it from the host OS) - but only if it's easy enough to feed `utc=False` into the logger initialization via the `**overrides` it supports.  If that's not simple to do, don't bother.

I'd like @ztucker4 to weigh in on the other ASpace-specific aspects of the program.

Thanks --Andy

python/merge_duplicate_containers_aspace.py
henry-r18: @akohler When I change to `max()` using the TC counts as-is, I'm finding that `create_time` doesn't order correctly. Julia wants the earliest creation date in the case of ties in the TC counts, in other words the `min()` value. I think that's why the LLM chose `min()` with the negative TC count, as it orders both the negative TC counts and the dates in one operation.

Here's a snippet illustration with tuples standing in for those evaluated by the lambda function:
```
min_input = [
    (-3, "2026-01-02T00:00:00Z"),
    (-3, "2026-01-01T00:00:00Z"),  # earliest create time--should be chosen
    (-1, "2026-01-03T00:00:00Z"),
]

max_input = [
    (3, "2026-01-02T00:00:00Z"),
    (3, "2026-01-01T00:00:00Z"),  # earliest create time--should be chosen
    (1, "2026-01-03T00:00:00Z"),
]

print(min(min_input))
print(max(max_input))
---
(-3, '2026-01-01T00:00:00Z') <- result from min() correctly gives earliest
(3, '2026-01-02T00:00:00Z')  <- result from max() takes more recent date between first two tuples
```

akohler: @henry-r18 You're right, my non-LLM-assisted brain hallucinated.  I think my approach would need an extra operation to `min()` the date while `max()`ing the count, which definitely is not an improvement.  I withdraw my suggestion on this one :)

python/merge_duplicate_containers_aspace.py
henry-r18: @akohler I'm checking with Julia regarding location data on top containers. Will address this once I hear back from her.

henry-r18: @akohler I confirmed with Julia that she does not want location data to stop a group being processed in production mode, but I did make it so that the count is incremented both during dry runs and in production. I sent Julia a few sample outputs and logs via Slack, and she's happy with the results, so I think this should be good to go.

I made one small change to how location data is logged, logging the URI ref to the location record, rather than the full dict returned by the API.

Re-requesting review now.

python/merge_duplicate_containers_aspace.py
```diff
@@ -117,28 +119,28 @@ def _resolve_aos_for_tcs(
...
+    :param list[dict] tcs: List of top container records.
     :return: True if any top container has location data, False otherwise.
     """
     for tc in tcs:
         locations = tc.get("container_locations", [])
         if locations:
             logger.warning(
-                f"Top container {tc.get('uri')} has location data: {locations}"
+                f"Top container {tc.get('uri')} has location data: "
+                f"{[location.get('ref') for location in locations]}"
```
akohler, line 133: This change triggers a test failure:
```
ERROR: test_check_for_location_data (test_merge_duplicate_containers_aspace.TestMergeDuplicateContainers.test_check_for_location_data)
Test that location data is flagged for review in logs if found.
----------------------------------------------------------------------
  File "/home/aspace/app/tests/test_merge_duplicate_containers_aspace.py", line 217, in test_check_for_location_data
    result = _has_location_data(test_tcs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aspace/app/merge_duplicate_containers_aspace.py", line 133, in _has_location_data
    f"{[location.get('ref') for location in locations]}"
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aspace/app/merge_duplicate_containers_aspace.py", line 133, in <listcomp>
    f"{[location.get('ref') for location in locations]}"
        ^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'get'
```

akohler (changes requested):
@henry-r18 The changes based on my previous review are good, except that there now is a test failing for me.  I rebuilt & restarted just in case I had old code, same result, so please take a look at that.

As noted in standup, the changes to the shared `configure_logging` method will need to be applied to other scripts using that method.  However, it didn't cause problems with my test run of  one of those, `find_duplicate_indicators.py`, so it's not a blocker here.  I'll open a new ticket for that cleanup as future work.

Once the failing test is fixed, this should be ready to merge.
Thanks --Andy

python/merge_duplicate_containers_aspace.py
henry-r18: Whoops! Sorry I missed that. Just pushed up a fix.

akohler (approved):
@henry-r18 👍 Merging now, thanks.

PR #28, opened 2026-06-25
DLSR-349: Script to migrate metadata from Alma to ASpace

ztucker4:
Implements [DLSR-349](https://uclalibrary.atlassian.net/browse/DLSR-349)

Adds a new script, `migrate_alma_metadata_to_archivesspace.py`, which updates ASpace Top Container records with metadata from matched Alma item records. This is intended as a companion to `add_alma_barcodes_to_archivesspace.py`, to be run after barcodes have been successfully transferred. It uses the same matching profiles and authentication setup. Full usage documentation is in the README under "Migrating Alma Metadata to ArchivesSpace."

Also includes updates to `utils/alma_utils.py` and `utils/aspace_utils.py`:

- `get_alma_items()` (wrapper for `get_alma_items_from_alma()` supporting caching) moved to `alma_utils.py`
- `_get_containers_from_container_refs()` and `get_aspace_containers()` moved to `aspace_utils.py`

These functions are taken from `add_alma_barcodes_to_archivesspace.py`, which should be updated to import them in a future ticket.

Pending before running in hosted ASpace instances:

LSC needs to decide how locations will be set up in ASpace. The script currently assumes that the "barcode" field on Locations will contain precisely the corresponding Alma location code. See the TODO in `_resolve_srlf_location_refs()`.

Example usage (using the LAUSD collection already barcoded using this matching profile):
```
docker compose exec python python migrate_alma_metadata_to_archivesspace.py \
    --bib_id 9969119863606533 \
    --holdings_id 22531983170006533 \
    --resource_id 1415 \
    --config_file .archivessnake_secret_DEV.yml \
    --dry_run \
    --print_output
```

[DLSR-349]: https://uclalibrary.atlassian.net/browse/DLSR-349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python/migrate_alma_metadata_to_archivesspace.py
```diff
@@ -0,0 +1,402 @@
...
+
+    :param ASnakeClient aspace_client: ASnakeClient instance.
+    :param set[str] srlf_codes: Alma location codes to resolve.
+    :return: Dict mapping Alma location code to ASpace location ref string.
+    """
+    logger.info("Resolving ASpace location refs for SRLF codes.")
+    resolved: dict[str, str] = {}
+
+    for location in aspace_client.get_paged("/locations"):
+        location_code = (
```
akohler: Why not just `location_code = location.get("barcode", "")`?

python/migrate_alma_metadata_to_archivesspace.py
```diff
@@ -0,0 +1,402 @@
...
+# DATA RETRIEVAL
+
+
+def _resolve_srlf_location_refs(
+    aspace_client: ASnakeClient, srlf_codes: set[str]
+) -> dict[str, str]:
+    """Build a mapping of SRLF location code to ASpace location ref at runtime.
+
+    Pages through ASpace locations to find entries matching the known SRLF codes,
+    returning early once all are resolved. Raises RuntimeError if any code cannot
```
akohler: Move the "Raises" info to a `:raises` to make it clearer to callers this can happen, and why.  As-is, it's harder to see in the general description, especially via tooltip when hovering over references elsewhere in code.

python/migrate_alma_metadata_to_archivesspace.py
```diff
@@ -0,0 +1,402 @@
...
+    """Set the container_profile ref on a top container dict based on VolEquiv.
+
+    Mutates tc in place. Returns True if the profile was applied, False if
+    skipped due to a missing or unrecognized VolEquiv value.
+
+    :param dict tc: ASpace top container dict to update.
+    :param dict alma_item: Alma item dict for the matched item.
+    :return: True if profile was applied, False otherwise.
+    """
+    raw = (alma_item.get("internal_note_1") or "").strip()
```
akohler: Use the default `.get(val, "")`

python/migrate_alma_metadata_to_archivesspace.py
```diff
@@ -0,0 +1,402 @@
...
+    if not raw:
+        logger.warning(
+            f"No VolEquiv (internal_note_1) found for item {alma_item.get('item_id')} "
+            f"(barcode {alma_item.get('barcode')}); skipping container profile update"
```
akohler, line 153: Is `alma_item` guaranteed to have a `barcode` here? `item_id` should be safe as it's system generated, but barcodes are optional on item records in Alma, and I do not see that `get_alma_items_from_alma()` ensures items have barcodes.

It looks like this routine is called in `main()` after other processing has confirmed a barcode will exist.  If so, just add a comment to confirm that these elements will exist, so readers know what conditions the callers have guaranteed will apply.

python/migrate_alma_metadata_to_archivesspace.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+    logger.info("Resolving ASpace location refs for SRLF codes.")
+    resolved: dict[str, str] = {}
+
+    for location in aspace_client.get_paged("/locations"):
+        location_code = (
+            location.get("barcode") or ""
+        )  # TODO: replace with correct field
+        if location_code in srlf_codes:
+            resolved[location_code] = location["uri"]
```
akohler, line 103: Does ASpace guarantee locations will have a `uri`?

python/migrate_alma_metadata_to_archivesspace.py
```diff
@@ -0,0 +1,402 @@
...
+    Mutates tc in place.
+
+    :param dict tc: ASpace top container dict to update.
+    :param dict alma_item: Alma item dict for the matched item.
+    """
+    logger.info(
+        f"Setting ILS Holding ID and ILS Item ID for top container {tc.get('uri')} "
+        f"from Alma item {alma_item.get('item_id')} (barcode {alma_item.get('barcode')})"
+    )
+    tc["ils_holding_id"] = alma_item.get("holding_id", "")
```
akohler: If Alma holding_id and item_id don't exist, should ASpace really be updated (later) with these empty values? As other comments show, I do like appropriate default values for `.get()` :) , but in this case missing values seem more like a data error than something to accept as-is.

python/migrate_alma_metadata_to_archivesspace.py
```diff
@@ -0,0 +1,402 @@
...
+    :param str timestamp: ISO 8601 timestamp string for the migration note.
+    """
+    logger.info(
+        f"Appending migration note to internal_note for top container {tc.get('uri')}"
+    )
+    migration_note = (
+        f"Metadata updated in a scripted batch migration "
+        f"from Alma item record on {timestamp}"
+    )
+    existing = (tc.get("internal_note") or "").strip()
```
akohler: Use the default `.get(val, "")`

python/migrate_alma_metadata_to_archivesspace.py
```diff
@@ -0,0 +1,402 @@
...
+        f"from Alma item {alma_item.get('item_id')} (barcode {alma_item.get('barcode')})"
+    )
+    tc["ils_holding_id"] = alma_item.get("holding_id", "")
+    tc["ils_item_id"] = alma_item.get("item_id", "")
+
+
+def _append_internal_note(tc: dict, timestamp: str) -> None:
+    """Append a migration note to the top container's internal_note field,
+    after a newline if there is existing content.
+
```
akohler, line 194: Like other methods, this should note that it mutates `tc` in place.

python/migrate_alma_metadata_to_archivesspace.py
```diff
@@ -0,0 +1,402 @@
...
+            "skipping location update"
+        )
+        return False
+
+    tc["container_locations"] = [
+        {
+            "jsonmodel_type": "container_location",
+            "ref": location_ref,
+            "status": "current",
+            "start_date": datetime.now(timezone.utc).strftime("%Y-%m-%d"),
```
akohler: This will vary as records are updated, so just confirming this is correct as opposed to the constant `timestamp` used for internal notes?

akohler (changes requested):
@ztucker4 The general flow of the script makes sense, with good documentation in the script and the README.

I flagged some small inconsistencies or places where extra clarification is helpful, both for now- and future-us.  When I don't see a default value for a `dict.get()` operation, and the code immediately afterwards doesn't check for existence of a value, I like to see either a meaningful default value or confirmation from the developer that there will always be a value.  I also prefer the more compact `.get(val, "")` to `(get(val) or "")`; [nice short discussion](https://stackoverflow.com/questions/33263623/dict-getkey-default-vs-dict-getkey-or-default), which does give performance reason to use the longer form when merited.

Finally, given the `SRLF` rename, let's switch to `SLFS`.  It's true the legacy Alma location codes remain `sr...` but everything else has changed, or will, and future maintainers that haven't lived through this change will appreciate one less thing to decode :)

Thanks --Andy

python/migrate_alma_metadata_to_archivesspace.py
```diff
@@ -0,0 +1,402 @@
...
+    :param dict unhandled_data: Unmatched/duplicate items and containers.
+    :param bool print_output: If True, also print to console.
+    """
+    summary_lines = [
+        f"Total Alma items: {len(alma_items)}",
+        f"Total ASpace top containers: {len(aspace_containers)}",
+        f"Matched top containers: {len(matched)}",
+        f"Location updates skipped (non-SRLF): {len(skipped_location)}",
+        f"Container profile updates skipped (missing/unrecognized VolEquiv): \
+            {len(skipped_profile)}",
```
henry-r18: I'd suggest using implicit string concatenation for these long lines, rather than backslashes, i.e.

```
(
    f"Container profile updates skipped (missing/unrecognized VolEquiv): "
    f"{len(skipped_profile)}",
)
```

henry-r18 (commented):
@ztucker4 Here's the output I got when running the example command you provided:

```
Total ASpace top containers: 2787
Matched top containers: 2787
Location updates skipped (non-SRLF): 0
Container profile updates skipped (missing/unrecognized VolEquiv):             0
Unmatched Alma items: 5
Unmatched ASpace top containers:             0
Alma items with duplicate keys:             0
ASpace top containers with duplicate keys:             0
```

One small suggestion for the `_print_summary` function is to use the implicit string concatenation style for the longer lines by wrapping them in parentheses, rather than using backslashes. I think that will prevent the tab-like spaces between the keys and values on the long lines.

python/migrate_alma_metadata_to_archivesspace.py
ztucker4: This comment revealed a bug - neither `item_id` nor `holding_id` is actually present in the Alma item record, so the empty string cases were always used here. `item_id` is actually called `pid`, which should always be present, so I've added a comment justifying the use of `alma_item["pid"]` here. For `ils_holding_id`, we can use the holding ID passed in at runtime, which is the same for all items in a collection.

python/migrate_alma_metadata_to_archivesspace.py
ztucker4: I think this was incorrect - I've updated the function to use the same timestamp here.

akohler (approved):
@ztucker4 This all looks good to me now, thanks for making the changes.  Merging now.

PR #30, opened 2026-07-24
DLSR-465: build and push Docker Hub image for ASpace scripts

ztucker4:
Implements [DLSR-465](https://uclalibrary.atlassian.net/browse/DLSR-465)

Adds `.github/workflows/build_docker_hub.yml` to build and publish the `python` service image to Docker Hub on every merge to main. I don't believe any other files need to be changed, since the existing Dockerfile already produces a self-contained image.

My work is based on the [two](https://github.com/UCLALibrary/ftva-lab-data/blob/main/.github/workflows/build_docker_hub.yml) [examples](https://github.com/UCLALibrary/ucla-ca-providence/blob/main/.github/workflows/build_docker_hub.yml) linked in the ticket.

Notes on my implementation:
* Both examples use a Chart for tagging with version numbers, but this one always uses `:latest`. The [ftva-lab-data example](https://github.com/UCLALibrary/ftva-lab-data/blob/main/.github/workflows/build_docker_hub.yml) also includes a `:scan` tag - I'm not sure what this is for, so I didn't include it.
* Both examples have `context: .`, but this one has `context: ./python` so we only package the script-running image.
* I've copied the Docker Hub secret references (`secrets.DOCKERHUB_USERNAME` and `secrets.DOCKERHUB_TOKEN`) directly from the examples - not sure how to confirm if these are correct.
* The extent of my testing has been to confirm that I can build the image with `docker build ./python -t as_test` and then run scripts, e.g. `docker run as_test python add_alma_barcodes_to_archivesspace.py [args...]`.

[DLSR-465]: https://uclalibrary.atlassian.net/browse/DLSR-465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


akohler (approved):
@ztucker4 Your action script looks correct.  I made a few changes to update the major versions of the actions used, via the [Marketplace](https://github.com/marketplace?type=actions).  This process needs improvement, someday.

Your use of `context` is great, including just what we need.  The approach to `label` also is correct - the `scan` tag/label is older, from when we used Snyk to do security scanning.

Pushing to Docker Hub the first time for a project requires extra setup in DH, which I did via our [internal documentation](https://uclalibrary.atlassian.net/wiki/x/d_T-Aw). This requires admin rights in DH, as a one-time process for each repository we push.  I also added the Docker secrets referenced in the script to this GH repository.

There's also no great way to test GA scripts.  After the previous setup, I copied yours and modified the actions so it would run immediately on push to any branch.  Did that, my [test run succeeded](https://github.com/UCLALibrary/archivesspace-toolkit/actions/runs/30114833270), so confidence is high.  I also pulled the built image down and confirmed that I can run & connect to it, and dependencies within it look correct.  I didn't try running any scripts, that can come with your testing on the new server :)

Finally, since I was updating stuff anyhow, I updated the devcontainer extensions to pin them & cleaned up the formatting.

Merging now, then I'll follow up with next steps in Jira.  Thanks --Andy

PR #31, opened 2026-07-31
DLSR-488: new shell script to coordinate running scripts via Docker on ASpace server

ztucker4:
Implements [DLSR-488](https://uclalibrary.atlassian.net/browse/DLSR-488)

Adds a new shell script, `run_aspace_script.sh`, intended for use on the ArchivesSpace support server. The script acts as a wrapper for the python scripts in this repo: it mounts all supplied secrets files, passes the script name and parameters to a `docker run` command, and uses `docker diff` to identify any new JSON or .log output files and copy them onto the server.

I have a couple of questions/things I'm not yet happy with:
* I'm not sure where we want to keep files on the support server going forward. My user only has read/write permissions within the `ztucker` directory, and this is currently hard-coded in the script as the parent directory for config and log files. I want to keep usage of the wrapper script as simple as possible, but maybe the directory names should be passed in as parameters or handled with environment variables.
* There isn't a great way to see output log files in real time, but there are a couple of workarounds. It is possible to connect to the server in a new terminal window, find the ID of the running container, and then use something like `docker exec -it aspace-run-12345 bash -c 'tail -f *.log'`. Alternatively, our most complex scripts (`add_alma_barcodes_to_archivesspace.py` and `migrate_alma_metadata_to_archivesspace.py`) have a `--print_output` argument that can be supplied when the script is initiated. I think these are probably sufficient to allow us/LSC staff to monitor processes, but it's still a bit clunky.

[DLSR-488]: https://uclalibrary.atlassian.net/browse/DLSR-488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


akohler (changes requested):
@ztucker4 I think using `docker compose`, with a simple `docker-compose_scripts.yml` or similar, is a better approach.  You could still have a shell script, but the compose file would specify any volume(s) needing mounting. 

The compose file would specify the image, maybe a "sleep infinity" command to keep a container running until explicitly stopped, and volume(s) if needed.  Just one volume should be enough, since the host directory being mounted can have `logs/` and `output/` and `secrets/` sub-directories as needed.

I've asked John about a shared space on the server, so we don't have to worry about individuals running scripts and then not being able to easily see / share files in their own home directories.  While waiting for that, you could work on the `docker compose` approach.

If you have questions please ask, I'm happy to help, especially if I've overlooked something.

run_aspace_script.sh
```diff
@@ -0,0 +1,38 @@
...
+# mounting in config files and copying out any log/output files it creates.
+
+# Be strict about errors, unset variables, and pipeline failures.
+set -euo pipefail
+
+# Define the Docker Compose file and data directory for use with ASpace scripts.
+# Use dirname and readlink to get the absolute path of the current script's directory,
+# and then append the docker-compose_scripts.yml filename to it.
+COMPOSE_FILE="$(dirname "$(readlink -f "$0")")/docker-compose_scripts.yml"
+export ASPACE_DATA_DIR="${ASPACE_DATA_DIR:-$HOME/aspace-data}"
```
akohler, line 12: Using `$HOME` here may give undesired results, if this runs anywhere except on the server.  For example, I just cleaned up new directories and a log file in my own home directory :) .  Not flagging this as something that needs changing, but if you're expecting consistency, you might want to define `ASPACE_DATA_DIR` in terms of the current directory (where you're finding `COMPOSE_FILE`).

run_aspace_script.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Check that the user provided at least one argument (the script name). 
+# If not, print usage information and exit with an error code.
+if [ "$#" -lt 1 ]; then
+  echo "Usage: $(basename "$0") <script_name.py> [script args...]"
+  exit 1
+fi
+
+# Bring up the Docker container in detached mode, so it runs in the background.
+docker compose -f "${COMPOSE_FILE}" up -d
```
akohler: This is more useful for systems which need to stay up and running, accepting connections, or have a long / complex startup.  Since this system will just run a given script, I'd probably not bring it `up` like this, but instead would `run` the script - see my next comment.

run_aspace_script.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Run the specified Python script inside the container, passing along any additional arguments. 
+# Capture the exit code for later use.
+set +e  
+docker compose -f "${COMPOSE_FILE}" exec scripts python "$@"
```
akohler: `exec` connects to a running service.  Since there's no complex startup needed, instead of bringing it `up`, `exec`-ing something, and then shutting it `down` - which this script does not do, it leaves the service running - I suggest just using `run` here instead.  Automatic startup and shutdown, though an explicit `down` can be useful to force the network created by `docker compose` to be removed too.

docker-compose_scripts.yml
```diff
@@ -0,0 +1,12 @@
+services:
+  scripts:
+    image: uclalibrary/archivesspace-toolkit:latest
+    container_name: aspace-scripts
+    command: sleep infinity
```
akohler: If you do switch to using `docker compose run` as suggested in my comments on the shell script, you may not need the `command` here.  It's needed when `exec`-ing the service, to keep it from exiting immediately if started via `up`, but I think won't be needed if you are just `run`-ing scripts.  Do test this though, I may be wrong.

docker-compose_scripts.yml
```diff
@@ -0,0 +1,12 @@
...
+    user: "${ASPACE_UID}:${ASPACE_GID}"
+    environment:
+      - HOME=/home/aspace
+    volumes:
```
akohler, line 7: I assume you're mounting these separate volumes so you can include just these 3 sub-directories into the container, and not whatever else is in the current directory (like this compose file)?  If so, that' s OK, and provides flexibility.  If there won't be much else in the current directory, you could simplify by mounting the current directory, or having `$ASPACE_DATA_DIR` be a sub-directory containing just the 3 sub-sub-directories.  Either is OK, no change requested.

akohler (changes requested):
@ztucker4 I think this is getting close.  My comments mainly suggest refining the use of `docker compose`, to run directly instead of starting up, attaching via `exec`, and then shutting down.  Either way, please make sure the service is shut down when the script ends.  You can check this interactively with `docker compose -f docker-compose_scripts.yml ps`  or just `docker ps` for a more general view.

Note that the compose file *requires* values provided by the script.  This shouldn't be a problem for Julia, but if someone does run this directly, they wind up as `root` in the container, and may be confused:
```
$ docker compose -f docker-compose_scripts.yml run scripts bash
WARN[0000] The "ASPACE_DATA_DIR" variable is not set. Defaulting to a blank string.
WARN[0000] The "ASPACE_DATA_DIR" variable is not set. Defaulting to a blank string.
WARN[0000] The "ASPACE_DATA_DIR" variable is not set. Defaulting to a blank string.
WARN[0000] The "ASPACE_UID" variable is not set. Defaulting to a blank string.
WARN[0000] The "ASPACE_GID" variable is not set. Defaulting to a blank string.
root@e8aa986d983b:~/app# exit
```

I didn't test beyond running the shell script with a valid python script name, because my local environment requires a tunneled connection to the hosted ASpace systems, and this compose file can't use that tunnel.  In the simpler target environment, already on the allowed IP list, that should all work fine.

README.md
```diff
@@ -336,3 +336,43 @@ ucla-kohler@aspace-hosting-production-bastion.lyrtech.org
 ```
 This tunnel runs in the foreground, so you'll need to open a second connection to run programs which need it.
 
+## Running scripts on the support server (UCLA only)
+
+For running scripts against hosted Test/Production ArchivesSpace outside of local development, a stand-alone image is published to Docker Hub as `uclalibrary/archivesspace-toolkit`. This is intended for use on the support server, `p-u-asrunner01.library.ucla.edu`, which is already on the allowed IP list for the hosted systems and doesn't require the tunnel setup described above.
+
+### Setup
+
+1. Pull the image: `docker pull uclalibrary/archivesspace-toolkit:latest`
```
akohler, line 345: Pulling explicitly isn't needed, since the compose script will ensure that all resources have been pulled / updated.

README.md
```diff
@@ -336,3 +336,43 @@ ucla-kohler@aspace-hosting-production-bastion.lyrtech.org
...
+2. Copy `docker-compose_scripts.yml` and `run_aspace_script.sh` onto the server, in the same directory, and make sure the script is executable (`chmod +x run_aspace_script.sh`).
```
akohler, line 346: Instead of individuals doing this, it's preferred to work in `/opt/aspace-shared/` .  The goal was/is for individuals not having to worry about nobody else having access to logs/output in their home directories.

However, I see extra work would be required to make all new files in `/opt/aspace-shared/` be writable by the `docker` group (which we're all in, on that server).  If Julia's really the only person who will be running scripts 90+% of the time on the server, going with the home directory route is fine for now.  If things change / expand / become problematic, we'll worry about them then.

akohler (approved):
@ztucker4 I think this is fine now.  It's more awkward to make this kind of thing fit into a shared environment with individual users than I'd hoped, but if there's really just one active user, I'm OK with it as-is.  If problems come up, we'll figure out better solutions, but this is ready for use.

You may need to work with Julia to get the 2 relevant files in place. Using `curl` to download the "raw" versions from GitHub is probably the easiest, or she can `scp` them from her local copy of the repository into her home directory.  Or you can put them into /`opt/aspace-shared/` and she can copy them from there.  I'm happy to help if any questions or problems.

Merged, thanks --Andy

PR #32, opened 2026-08-05
DLSR-524: use consistent output directories for aspace scripts

ztucker4:
Implements [DLSR-524](https://uclalibrary.atlassian.net/browse/DLSR-524)

This should be merged before (or simultaneously with) https://github.com/UCLALibrary/archivesspace-toolkit/pull/31 ([DLSR-488](https://uclalibrary.atlassian.net/browse/DLSR-488))

Updates all Python scripts in this repo to consistently use two directories for output: `\logs` (for .log files) and `\output` (for everything else, currently JSON/txt reports and data cache files).

The bulk of the change is in `utils/generic_utils.py`. This file now defines `OUTPUT_DIR` and `LOGS_DIR`, and methods that write output now use these values as destination folders. Since most scripts in this repo import their output methods from this utils module, this fixed the majority of the scripts.

Only three scripts needed actual updates:
* `find_missing_containers_aspace.py`: updated default path, removed redundant `Path()` casting, and fixed info in print statements.
* `get_unlinked_top_containers.py`: as before, this script writes its own CSV without importing from `utils`. It now uses the new `resolve_output_path` to determine its destination.
* `get_container_counts.py`: removed redundant `Path()` casting, and fixed print statement.

The README is also updated in a few places to reflect the changed output destinations, and to fix an unrelated incorrect sample command.

[DLSR-524]: https://uclalibrary.atlassian.net/browse/DLSR-524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DLSR-488]: https://uclalibrary.atlassian.net/browse/DLSR-488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ docker compose exec python python add_alma_barcodes_to_archivesspace.py \
     --bib_id 123456789 \
     --holdings_id 987654321 \
     --resource_id 1234 \
-    --profile config.indicator_type_matching.py \
+    --profile config.indicator_type_matching \
     --config_file .archivessnake_secret_DEV.yml \
     --dry_run \
     --use_cache \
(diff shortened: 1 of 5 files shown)


python/utils/generic_utils.py
```diff
@@ -8,6 +8,25 @@
 from datetime import datetime
 from pathlib import Path
 
+# Shared output directories. These are expected to be mounted as volumes
+# (e.g. via `docker compose`) so that logs, reports, and cache files land
+# in predictable, host-accessible locations regardless of which script
+# or container produced them.
+LOGS_DIR = Path("/logs")
```
akohler: I'm concerned about these, because they're absolute paths (starting with `/`).  As-is, these will use `/logs` and `/output` within the container, which probably will require root access.

Instead, these probably should be `logs` and `output`, or maybe `./logs` and `./output` to look for them in the current working directory from python's POV.

README.md
```diff
@@ -185,10 +185,10 @@ We have only barcoded a few collections so far, so the profiles may need to be a
 ### Evaluating the script output
 
 After running the script, you will see several output files:
-- `add_alma_barcodes_to_archivesspace_{timestamp}.log`: The main log file.
-- `unhandled_add_alma_barcodes_to_archivesspace_{timestamp}.json`: A JSON file containing Alma items that were not matched to ArchivesSpace top containers.
-- `aspace_data_{resource_id}.json`: A JSON file containing the ArchivesSpace top container data for the resource. This file is used as a data source for the script if the `--use_cache` flag is set.
-- `alma_data_{bib_id}_{holdings_id}.json`: A JSON file containing the Alma item data for the collection. This file is used as a data source for the script if the `--use_cache` flag is set.
+- `/logs/add_alma_barcodes_to_archivesspace_{timestamp}.log`: The main log file.
```
akohler: Per my other comment on the values used, these probably should not have leading slashes.  `logs/abcde` tells us to look in a `logs` sub-directory of the current directory.  Could also be `./logs/abcde` with a `./` but that's less clear.  `/logs/abcde` means to look starting in the root directory.

akohler (changes requested):
@ztucker4 I think the only problem is the path designations.  I flagged a couple of uses, but if you do switch from `/logs` to `logs` etc., references in doc-strings probably should be changed for consistency.

akohler (approved):
@ztucker4 Looks good now, merging.

PR #33, opened 2026-08-13
DLSR-530: report missing Alma items (in addition to missing ASpace TCs)

ztucker4:
Implements [DLSR-530](https://uclalibrary.atlassian.net/browse/DLSR-530)

Previously, the script `find_missing_containers_aspace.py` was used to identify items in Alma that did not have a corresponding ASpace top container. LSC wants to also have the reverse comparison - top containers with no matching Alma item. 

As they will likely always want to check both directions simultaneously, this PR renames the old script to `reconcile_containers_bibs.py` and outputs both reports as separate CSV files.

This script uses a database connection to retrieve ASpace data, and thus can only be run from the ExL support server at the moment (after setting up the DB tunnel). I've pulled this branch there, and this example report can be run after activating the venv in the `python` directory:

```
python reconcile_containers_bibs.py \
-c .archivessnake_secret_PROD.yml \
-r 3915 \
--bib_id 9942304393606533 \
--holdings_id 22315350700006533
```

Example output has been sent to Julia for LSC's review.

[DLSR-530]: https://uclalibrary.atlassian.net/browse/DLSR-530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


akohler (changes requested):
@ztucker4 From my POV this looks pretty good, though I'm getting close to pushing for some custom classes or types to reduce the complexity of `tuple[dict[tuple, dict], list[tuple]]` type hints... but not today (unless you really want to).

I did flag one type mismatch, where the checker doesn't know what will be in `containers_by_uri`, which should be easy to fix.  Then I'll merge.

akohler (approved):
@ztucker4 All good now, merging.

PR #34, opened 2026-08-14
DLSR-545: Enable production db access on support server

akohler:
Implements [DLSR-545](https://uclalibrary.atlassian.net/browse/DLSR-545).

This PR modifies `run_aspace_script.sh` to tunnel a connection via SSH to the remote production database server.  This is required for several of our Python scripts.  The connection is automatically closed after Docker runs the Python script.

Notes and caveats:
* Lyrasis only supports access to our PRODUCTION database.  If Python scripts are being run against the TEST ASpace system, they may see a mixture of data from TEST (via API) and PRODUCTION (via database).  If this is a problem, we'll need to solve it, either via a flag to this shell script, or awareness within the Python scripts to not try database access when running against the TEST system.
* Users must have `id_aspace_ssh` in `~/.ssh/`, aka `$HOME/.ssh/`.  For now, this file can be copied from `/opt/aspace-shared/secrets/` on the `p-u-asrunner01` server.  This file must also have restrictive permissions, readable only by the owner.  As needed:
```
cp /opt/aspace-shared/secrets/id_aspace_ssh ~/.ssh/
chmod 600 ~/.ssh/id_aspace_ssh

# This  should now show the file, owned by the user, with -rw-------. permissions
ls -al ~/.ssh/id_aspace_ssh
```
* A change is required to each user's `secrets/.archivessnake_secret_PROD.yml`, on the `p-u-asrunner01` server:
```
database:
  host: 127.0.0.1
```
* It's possible this will not work correctly for a user, if another user is currently running a script.  Port 3306 on the host (the server) is being forwarded to the remote system, and if user A has already done that, I don't know what will happen if user B then tries to do that.  And if the forwarded port is shared, I think user A's data from the db might come back to user B's program.  We probably can fix this by picking a random local port to forward.

Testing:
* Pull down the new versions of `docker-compose_scripts.yml` and `run_aspace_script.sh` to your home directory on `p-u-asrunner01`.
* Make sure you have `~/.ssh/id_aspace_ssh` in place with correct permissions.
* Update your `secrets/.archivessnake_secret_PROD.yml` with the database host change.
* Use the updated shell script to run a program which interacts with the production database.

Bonus points if you can coordinate a multi-user test which will be safe if database data is returned to the wrong user's program.  We may need to write a testing Python program and add it to the container.

I added a warning to the end of the README.  Feel free to add other info as needed.

I added copious comments around my changes to the other files.  Please let me know if anything is unclear.

[DLSR-545]: https://uclalibrary.atlassian.net/browse/DLSR-545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ztucker4 (approved):
All looks good - see comment for notes on simultaneous testing. Thanks for the thorough documentation (especially the comments in `run_aspace_script.sh`). Merging!

